PR #75, opened 2018-05-24
 start services with docker-compose

graphaelli:
Adds `scripts/compose.py`, a utility for generating a `docker-compose.yml` useful for development and integration environments.

### DIFF
--- a/scripts/ci/go.sh
+++ b/scripts/ci/go.sh
@@ -4,6 +4,6 @@ srcdir=`dirname $0`
 test -z "$srcdir" && srcdir=.
 . ${srcdir}/common.sh
 
-DEFAULT_COMPOSE_ARGS="master --with-agent-go-net-http --force-build"
+DEFAULT_COMPOSE_ARGS="master --no-kibana --with-agent-go-net-http --force-build"
 export COMPOSE_ARGS=${COMPOSE_ARGS:-${DEFAULT_COMPOSE_ARGS}}
 runTests env-agent-go docker-test-agent-go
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -6,19 +6,109 @@ __WIP__
...
+
+Topics are named according to service name. To view events for 1234_service-12a3:
+
+    docker exec -it localtesting_6.3.0-SNAPSHOT_kafka kafka-console-consumer --bootstrap-server kafka:9092 --topic apm-1234_service-12a3 --from-beginning --max-messages 100
+
+#### Advanced topics
+
+##### Dumping docker-compose.yml
+
+    `start --docker-compose-path --skip-download -` will dump the generated `docker-compose.yml` to standard out without starting any containers or downloading images.  Omit `--skip-download` to just download images.
```
jalvz: can you put an example of specific command? how exactly do i use this?

Makefile
```diff
@@ -1,10 +1,82 @@
...
+	docker-compose down -v --remove-orphans || true
+
+destroy-env: venv
+	(docker ps -aqf network=apm-integration-testing | xargs docker rm -f && docker network rm apm-integration-testing) || true
+
+# default (all) started for now
+env-%: venv
+	$(MAKE) start-env
+
+test: test-all lint
```
jalvz: would be nice to document briefly all these targets: what do they do, what depends on what, if i am meant to run something manually or they are only used internally, etc.

README.md
simitt: I could dump it with `./scripts/compose.py start --docker-compose-path apmtesting.yml 6.4 --skip-download`

graphaelli:
closing, since opening this repo I lost access to update this pull request.

README.md
jalvz: thanks! then, it doesn't dump to stdout but to a file passed as argument?

simitt: yes, I haven't tried to dump to stdout

PR #82, opened 2018-06-05
kill any running containers before starting under jenkins

graphaelli:
instead of when not running under jenkins


graphaelli:
no, it's already right

Issue #80, opened 2018-06-05
make force build skip docker cache


graphaelli:
While using the docker cache changes in agent repos do not trigger container image rebuilds.

`--force-build` currently adds the `--build` option to `docker-compose up` - that does not skip the docker cache.  Add an extra step to `docker-compose build --no-cache` when `--force-build` is enabled.

graphaelli:
closed by #113

PR #85, opened 2018-06-06
use `docker-compose build` to build images instead of `--build` flag

beniwohli:
When using the `--build` flag, docker-compose will only pull images if
it doesn't have a matching image locally. But since we update the
opbeans images semi-often, we want to ensure that the newest version
is always pulled. This is what `docker-compose build --pull` does.

@graphaelli any reason we don't _always_ build the images? That's what we did in localtesting. With the layer cache, it takes only a second or two if everything is current, and I suspect it would save us from countless hours of people wondering and trying to debug why they don't have the current version of the different opbeans variations running.

### DIFF
--- a/scripts/compose.py
+++ b/scripts/compose.py
@@ -1806,7 +1806,7 @@ class LocalSetup(object):
             print("Starting stack services..\n")
             cmd = ['docker-compose', '-f', docker_compose_path.name, 'up', '-d']
             if args["force_build"]:
-                cmd.append("--build")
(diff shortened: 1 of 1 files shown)


graphaelli:
no reason, I was thinking of it like #80 but if people are already used to always building that sounds good to me.

beniwohli:
In that case I'd change the behaviour to always run `docker-compose build --pull` and, if `--force-build` is provided, use the `--no-cache` flag

graphaelli:
superseded by #113

PR #86, opened 2018-06-06
update opbeans-python localtesting Dockerfile to use upstream base image

beniwohli:


graphaelli:
The build failure is unrelated to this change and was resolved in a separate pull request, good to merge.

beniwohli:
Thanks for the clarification Gil, merging :)

PR #92, opened 2018-06-11
remove runtime dependency on PyYAML

graphaelli:
still needed for tests, for now

### DIFF
--- a/scripts/compose.py
+++ b/scripts/compose.py
@@ -31,8 +31,11 @@ import sys
 import subprocess
 import unittest, unittest.mock
 
-import yaml
-
+# TODO: convert yaml fixtures and remove this, only needed for tests
+try:
+    import yaml
+except:
+    pass
 #
 # package info
 #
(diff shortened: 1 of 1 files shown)


graphaelli:
there are some other deps to remove before the venv isn't required, will follow up with those later.

scripts/compose.py
```diff
@@ -1813,14 +1816,13 @@ def start_handler(self):
                 pgdata={"driver": "local"},
             ),
         )
-        yaml_opts = dict(
-            explicit_start=True,
-        )
         docker_compose_path = args["docker_compose_path"]
-        yaml.dump(compose, docker_compose_path, **yaml_opts)
+        json.dump(compose, docker_compose_path)
```
beniwohli: do you think it would make sense to change the default value for `docker_compose_path` to `docker-compose.json`?

https://github.com/elastic/apm-integration-testing/blob/f51456e53b1045afe96633d6eecd3de178d27afb/scripts/compose.py#L1629

PR #94, opened 2018-06-12
Ruby agent return filtered payload.

simitt:
When adding a filter for the ruby agent,
return the filtered payload,
instead of the rejected elements.


mikker:
👍🏻 

Also: https://github.com/elastic/apm-agent-ruby/pull/151

PR #95, opened 2018-06-12
generate junit results

graphaelli:
for CI to make sensible failure messages and reports

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,9 @@ VENV ?= ./venv
 
 COMPOSE_ARGS ?=
 
+JUNIT_RESULTS_DIR=tests/results
+JUNIT_OPT=--junitxml $(JUNIT_RESULTS_DIR)
+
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -66,6 +69,8 @@ dockerized-test:
 	@echo running make $(TARGET) inside a container
 	docker build --pull -t apm-integration-testing .
 
+	mkdir -p -m 777 "$(PWD)/$(JUNIT_RESULTS_DIR)"
+	chmod 777 "$(PWD)/$(JUNIT_RESULTS_DIR)"
```
beniwohli: Out of interest, why the `chmod 777` after you already created the directory with `-m 777`?

graphaelli: Thanks for the speedy review turnaround.

The `mkdir` didn't change perms if the directory already existed, the case when I tested locally.  I forgot to remove it after adding the `chmod` but I don't suppose it hurts.

PR #100, opened 2018-06-14
Add opbeans java

alvarolobato:


graphaelli:
Can you add a test under `OpbeansServiceTest`? Nothing fancy needed, the intention is to protect against regressions.

scripts/compose.py
```diff
@@ -868,6 +868,69 @@ def _content(self):
...
+        self.agent_branch = options.get("opbeans_agent_branch", self.DEFAULT_AGENT_BRANCH)
+        self.agent_repo = options.get("opbeans_agent_repo", self.DEFAULT_AGENT_REPO)
+
+    def _content(self):
+        depends_on = {
+            "elasticsearch": {"condition": "service_healthy"},
+            "postgres": {"condition": "service_healthy"}
+        }
+
+        if self.options.get("with_apm-server", True):
```
graphaelli: I see we do this in `OpbeansNode` too - it should be `if not self.options.get("disable_apm_server", False)`

graphaelli: And that's merged now, you probably want to rebase, you'll have a failing test that point.

alvarolobato: @graphaelli fixed and rebased

scripts/compose.py
```diff
@@ -886,6 +886,69 @@ def _content(self):
...
 
+class OpbeansJava(Service):
+    SERVICE_PORT = 3002
+    DEFAULT_AGENT_BRANCH = "master"
+    DEFAULT_AGENT_REPO = "elastic/apm-agent-java"
+    DEFAULT_LOCAL_REPO = "."
+
+    @classmethod
+    def add_arguments(cls, parser):
+        super().add_arguments(parser)
```
graphaelli: we've decided to make things mostly work for python2 - can you change this to `super(OpbeansJava, self).__init__(**options)`

scripts/compose.py
```diff
@@ -1441,7 +1504,53 @@ def test_opbeans_ruby(self):
...
+                      postgres:
+                        condition: service_healthy
+                      apm-server:
+                        condition: service_healthy
+                    volumes:
+                      - .:/local-install
+                    healthcheck:
+                      test: ["CMD", "curl", "--write-out", "'HTTP %{http_code}'", "--silent", "--output", "/dev/null", "http://opbeans-java:3002/"]
+                      interval: 5s
+                      retries: 12""")
```
graphaelli: we don't care about line length in this text block, you can add `# noqa: 501` at the end of 1552

scripts/compose.py
alvarolobato: ha, sorry I didn't get the notification.  Fixing it.

graphaelli:
lgtm when you're done linting

alvarolobato:
I'm going to end up learning python with this one 😜 ... should work now

graphaelli:
looks like it's really really fixed now

alvarolobato:
Yeah, I installed the python tools and ran it locally

PR #102, opened 2018-06-17
docs: fix headline levels

watson:
It was a bit disorienting for me when reading through the README the first time. I think maybe some of the headlines had been moved around and didn't have the correct level anymore. I've tried my best to sort it out, but let me know if I misunderstood the intentions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,13 +18,13 @@ brew install pyenv-virtualenv
 brew cask install docker
 ```
 
-### Starting an Environment
+## Starting an Environment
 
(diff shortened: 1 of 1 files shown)


graphaelli:
and now I broke it, sorry

watson:
I fixed the conflict by re-indenting the two `Running Local Enviroments` and `Stopping Local Enviroments` headlines. But I'm now not sure if the examples should be under the new `Running Local Enviroments` headline as well or if that headline is actually needed with the changes introduced by the PR?

PR #99, opened 2018-06-14
compose.py python2.7 support

graphaelli:
not hooked up to CI, keeping it best effort for now

PR #103, opened 2018-06-17
docs: update prerequisite section

watson:
Renamed "Installation Requirements" to "Prerequisites" and attempted to make it more universal and less reliant on non-standard package managers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ This repo is tested with Python 3 but best effort is made to make starting/stopp
   ```sh
   brew install python
   ```
-- Linux
+- Debian/Ubuntu
   ```sh
   sudo apt-get install python3
   ```
--- a/README.md
+++ b/README.md
@@ -4,20 +4,30 @@ This repo contains tools for end to end (eg agent -> apm server -> elasticsearch
 
 [![Build Status](https://apm-ci.elastic.co/view/All/job/elastic+apm-integration-testing+master+push/badge/icon?style=plastic)](https://apm-ci.elastic.co/job/elastic+apm-integration-testing+master+push/)
 
-## Installation Requirements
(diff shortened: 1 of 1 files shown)


graphaelli:
This shouldn't be needed once #99 goes in.   That in turn was blocked but I think we can resolve all of that today.

watson:
It should be updated to show support for Python 2 sure, but what do you mean by it's not needed?

watson:
Ah, is the virtualenv requirement gone as well?

graphaelli:
Right, no need to virtualenv for just standing up an environment, only python 2 or 3 stdlib is used. 
 virtualenv + python3 is required for running the apm integration tests, even under docker.

graphaelli:
@watson good to close this one?

watson:
Lost track of it. I still think the requirements section could be improved a bit, so I tried to address your earlier comments and rebased with master.

I wanted to get the reference to installing docker with Homebrew out and instead link to the official Docker install docs. If there was a reason why we showed the brew instructions let me know 😃 

I also added a section on how to install Python (but left the note about the Python versions)

What do you think?

README.md
```diff
@@ -4,20 +4,30 @@ This repo contains tools for end to end (eg agent -> apm server -> elasticsearch
...
+[Installation instructions](https://www.docker.com/community-edition)
+
+### Python 3
+
+- Windows: [Installation instructions](https://www.python.org/downloads/windows/)
+- Mac (using [Homebrew](https://brew.sh/)):
+  ```sh
+  brew install python
+  ```
+- Linux
```
graphaelli: This should be something like `Debian/Ubuntu`

watson: Good point. Fixed :)

PR #110, opened 2018-06-25
add opbeans real user monitoring container

graphaelli:
image

Missed in the unification from localtesting.

cc @formgeist @jahtalab


formgeist:
@graphaelli Thanks for the quick turn-around on getting this in as well 👍

makwarth:
sweet!

PR #114, opened 2018-06-26
add --elasticsearch-plugin option

graphaelli:


graphaelli:
No longer have a need for this.

PR #115, opened 2018-06-27
Add opbeans-go service

axw:
Add an opbeans-go service to compose.py, using both Postgres and Redis. Like opbeans-python, opbeans-go reuses the opbeans-frontend docker image.

It is not currently possible to specify a local Go agent repo.

### DIFF
--- a/docker/opbeans/go/Dockerfile
+++ b/docker/opbeans/go/Dockerfile
@@ -1,0 +1,23 @@
+FROM golang:1.10
+ARG GO_AGENT_BRANCH=master
+ARG GO_AGENT_REPO=elastic/apm-agent-go
+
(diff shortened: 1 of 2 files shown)


axw:
> This is great! I think this is the first opbeans app that reports itself and the frontend agent - very nice.

opbeans-python has support for this too, although I'm not sure if it works OOTB.

> Two ideas for the future:
>
> Monitor opbeans-go with metricbeat using the go module
> Log in json for easy filebeat parsing.

Sounds good, I'll create issues for these on opbeans-go.

axw:
> Monitor opbeans-go with metricbeat using the go module

Not sure about this one though. I think we should exercise the builtin metrics instead, when available.

PR #118, opened 2018-06-29
healthcheck as postgres user

graphaelli:
instead of root user, which worked but triggered these logs with each invocation:
```
[244] FATAL:  password authentication failed for user "root"
[244] DETAIL:  Role "root" does not exist.
Connection matched pg_hba.conf line 95: "host all all all md5"
```


graphaelli:
fun fact: discovered this issue after hooking up filebeat

PR #131, opened 2018-07-16
Add java spring integration tests

jalvz:


jalvz:
@felixbarny i am adding java integration tests here, is possible that there might be an issue with the Java agent, it would be great if you can check it out and have a look.

I can give you a pointer to get started, let me know if you have any questions:

To execute:
`./scripts/compose.py start master  --with-agent-java`
`make test-agent-java`

The tests are the same for all the agents. 
`test_req_java_spring` passes, and `test_concurrent_req_java_spring` and `test_concurrent_req_java_spring_foobar` fail.

The `test_concurrent_req-*` tests uses an async web framework/library called Tornado, and what we do there is to send a ton of requests simultaneously to the instrumented app, this is done here:
https://github.com/elastic/apm-integration-testing/blob/master/tests/agent/concurrent_requests.py#L93-L102
It is the same code for all the agents.

Then we wait until all the docs have been indexed into elasticsearch (we know how many requests we send, so we also know how many docs to expect) and run some assertions.
The java tests fail with a timeout error because the doc count stays at 0; ie. no data is sent.

Sending requests 1 by one with 1 second sleep between them; does work. 

You can eg. run `curl localhost:8090/foo` and then `curl localhost:9200/apm-7.0.0-alpha1-transaction-2018.07.18/_search` to validate.

If you have other means to send ~1k requests in a few seconds, you can try that as well.

docker/java/spring/testapp/src/main/java/hello/GreetingController.java
```diff
@@ -0,0 +1,41 @@
...
+    @GetMapping("/")
+    @ResponseBody
+    public String index() {
+        return "OK";
+    }
+
+    @GetMapping("/foo")
+    @ResponseBody
+    public String foo() {
+        Span sp = ElasticApm.startSpan();
```
felixbarny: The API has just been changed, that's why the build errors. Use
```java
Span sp = ElasticApm.currentSpan().createSpan();
```
instead.

docker/java/spring/testapp/src/main/java/hello/GreetingController.java
```diff
@@ -0,0 +1,41 @@
...
+        sp.setType("just a regular foo");
+        String foo = new FooBar().foo();
+        sp.end();
+        return foo;
+    }
+
+    @GetMapping("/bar")
+    @ResponseBody
+    public String bar() {
+        Span sp = ElasticApm.startSpan();
```
felixbarny: same here

scripts/compose.py
```diff
@@ -778,10 +778,28 @@ def _content(self):
         )
 
 
+class AgentJavaSpring(Service):
+    SERVICE_PORT = 8090
+
+    def _content(self):
+        return dict(
```
graphaelli: can we get a healthcheck on this service to avoid flakiness in tests running before this is actually ready?

docker/java/spring/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+RUN mkdir /agent \
+    && mkdir /app
+
+COPY testapp /app
+
+WORKDIR /agent
+
+RUN curl -L https://github.com/elastic/apm-agent-java/archive/master.tar.gz | tar --strip-components=1 --exclude=demo/db.sql.bz2 -xzv \
+    && mvn package -DskipTests \
```
felixbarny: add `--batch-mode`
This prevents the download progress indicators which blow up the logs considerably

felixbarny: Done

docker/java/spring/testapp/src/main/java/hello/GreetingController.java
```diff
@@ -0,0 +1,41 @@
...
+        sp.setType("just a regular foo");
+        String foo = new FooBar().foo();
+        sp.end();
+        return foo;
+    }
+
+    @GetMapping("/bar")
+    @ResponseBody
+    public String bar() {
+        ElasticApm.currentSpan().createSpan();
```
graphaelli: missed setting `sp` here

felixbarny: I'm on it

docker/java/spring/testapp/src/main/java/hello/FooBar.java
```diff
@@ -0,0 +1,25 @@
...
+import co.elastic.apm.api.Span;
+
+public class FooBar {
+
+    public String foo() {
+        return "foo";
+    }
+
+    public String bar() {
+        ElasticApm.currentSpan().createSpan();
```
graphaelli: missed setting sp here

docker/java/spring/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+    && mv elastic-apm-agent/target/elastic-apm-agent-${JAVA_AGENT_BUILT_VERSION}.jar apm-agent.jar \
+    && mvn --batch-mode install:install-file -Dfile=apm-agent-api/target/apm-agent-api-${JAVA_AGENT_BUILT_VERSION}.jar -DgroupId=co.elastic.apm -DartifactId=apm-agent-api -Dversion=${JAVA_AGENT_BUILT_VERSION} -Dpackaging=jar \
+    && cd /app \
+    && mvn --batch-mode -DAGENT_API_VERSION=${JAVA_AGENT_BUILT_VERSION} package
+
+WORKDIR /app
+
+EXPOSE 8090
+
+CMD ["java", "-javaagent:../agent/apm-agent.jar", "-Delastic.apm.service_name=my-service", "-Delastic.apm.application_packages=hello", "-jar","target/hello-spring-0.1.jar"]
```
graphaelli: `s/my-service/springapp/`

felixbarny:
Need help with decrypting the logs. Not sure what causes the failure.

jalvz:
I fixed some assertions, works locally for me now. 
CI logs point to networking issues, looking into it...

scripts/compose.py
```diff
@@ -792,6 +792,7 @@ def _content(self):
                 "ELASTIC_APM_SERVICE_NAME": "springapp",
                 "ELASTIC_APM_SERVER_URL": "http://apm-server:8200",
             },
+            healthcheck=curl_healthcheck(self.SERVICE_PORT, "javaspring", path="/"),
```
felixbarny: the health check url is `/healthcheck`. `/` is traced

jalvz:
jenkins retest this please

Makefile
```diff
@@ -88,6 +88,7 @@ dockerized-test:
 	  -e EXPRESS_URL="http://expressapp:8010" \
 	  -e FLASK_URL="http://flaskapp:8001" \
 	  -e GO_NETHTTP_URL="http://gonethttpapp:8080" \
+	  -e JAVA_SPRING_URL="http://javaspring:8090" \
```
jalvz: gazillion thanks @graphaelli . owned 🍻 to you ++

felixbarny:
It also seems like it does not make the build considerably slower. It takes about 10 min now vs 8 min previously.

PR #139, opened 2018-07-24
hardcode internal port of opbeans-python to 3000

beniwohli:
This is the first step in harmonizing the internal ports of all opbeans images. The goal is that all opbeans containers are reachable under `opbeans-XYZ:3000`


scripts/compose.py
```diff
@@ -1063,7 +1063,7 @@ def _content(self):
             image=None,
             labels=None,
             healthcheck=curl_healthcheck(self.SERVICE_PORT, "opbeans-python", path="/"),
-            ports=[self.publish_port(self.port, self.SERVICE_PORT)],
+            ports=[self.publish_port(self.port, 3000)],
```
graphaelli: why this rather than changing `OpbeansPython.SERVICE_PORT` ?

graphaelli: oh I see, it's still 8000 to the host, sorry

PR #141, opened 2018-07-24
fix: pin pm2 i version 2

watson:
Version 3 have a bug that means it can't start


watson:
For reference, this was the issue in pm2: https://github.com/Unitech/pm2/pull/3826

PR #143, opened 2018-07-25
bump version

jalvz:

PR #147, opened 2018-07-26
 test with 6.4 stack snapshots

graphaelli:
and bump versions for the rest of 6.x

### DIFF
--- a/scripts/compose.py
+++ b/scripts/compose.py
@@ -55,7 +55,7 @@ else:
 PACKAGE_NAME = 'localmanager'
 __version__ = "4.0.0"
 
-DEFAULT_STACK_VERSION = "6.3.1"
+DEFAULT_STACK_VERSION = "6.3.3"
 
 
 #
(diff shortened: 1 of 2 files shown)


scripts/compose.py
```diff
@@ -55,7 +55,7 @@ def patch(_):
 PACKAGE_NAME = 'localmanager'
 __version__ = "4.0.0"
 
-DEFAULT_STACK_VERSION = "6.3.1"
+DEFAULT_STACK_VERSION = "6.3.3"
```
jalvz: i opened https://github.com/elastic/apm-integration-testing/pull/143 they other and then forgot about it.
why 6.3.3?

graphaelli: Since those are the latest snapshots, 6.3.2 snapshots will go away soon enough.  We can go back to 6.3.2 when 6.3 releases stop for good, if that's the last release.

jalvz: right. we might need different things in the future, eg. for non-apm devs friendliness `--release` should be maybe default and the `DEFAULT_STACK_VERSION` point to the latest release.
for now this is ok, will close 143 in favor of this 👍

PR #148, opened 2018-07-26
Add RUM JS integration test

jahtalab:
cc @roncohen @alvarolobato

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,7 +168,7 @@ For example, to test an update to the python agent from user `elasticcontributor
 ```bash
 # start test deps: apm-server, elasticsearch, and the two python test applications
 # the test applications will use elasticcontributor's newfeature1 apm agent
-./script/compose.py start master --no-kibana --with-agent-python-django --with-agent-python-flask --python-agent-package=git+https://github.com/elasticcontributor/apm-agent-python.git@newfeature1 --force-build
+./scripts/compose.py start master --no-kibana --with-agent-python-django --with-agent-python-flask --python-agent-package=git+https://github.com/elasticcontributor/apm-agent-python.git@newfeature1 --force-build
 
 # wait for healthiness
 docker-compose ps
--- a/tests/agent/test_rum.py
+++ b/tests/agent/test_rum.py
@@ -6,8 +6,8 @@ import requests
 def test_rum(rum):
     elasticsearch = rum.apm_server.elasticsearch
     elasticsearch.clean()
-    endpoint = Endpoint(rum.url, "run_integration_test")
+    endpoint = Endpoint(rum.url, "run_integration_test", qu_str="echo=done", text="done")
 
     r = requests.get(endpoint.url)
-    assert r.status_code == 200
+    utils.check_request_response(r, endpoint)
     utils.check_elasticsearch_transaction(elasticsearch, 1, {'query': {'term': {'processor.event': 'transaction'}}})
--- a/tests/agent/test_rum.py
+++ b/tests/agent/test_rum.py
@@ -1,0 +1,13 @@
+from tests.endpoint import Endpoint
+from tests import utils
+import requests
+
+
+def test_rum(rum):
+    elasticsearch = rum.apm_server.elasticsearch
+    elasticsearch.clean()
+    endpoint = Endpoint(rum.url, "run_integration_test")
+
+    r = requests.get(endpoint.url)
+    assert r.status_code == 200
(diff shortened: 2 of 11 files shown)


tests/utils.py
```diff
@@ -24,14 +24,15 @@ def check_request_response(req, endpoint):
     assert req.status_code == endpoint.status_code
 
 
-def check_elasticsearch_transaction(elasticsearch, expected_count):
-    q = {'query': {'term': {'processor.name': 'transaction'}}}
+def check_elasticsearch_transaction(elasticsearch,
+                                    expected_count,
+                                    query={'query': {'term': {'processor.name': 'transaction'}}}):
```
beniwohli: default arguments for functions/methods should never be mutable objects in Python, as they are initialized when the module is imported. e.g. imagine if you do `query['query'] = something_else` in the function body, you just changed the default for all following invocations of the function.

A better solution for this is:

    def check_elasticsearch_transaction(elasticsearch, expected_count, query=None):
        if query is None:
            query = {'query': {'term': {'processor.name': 'transaction'}}}

README.md
```diff
@@ -168,7 +168,7 @@ For example, to test an update to the python agent from user `elasticcontributor
 ```bash
 # start test deps: apm-server, elasticsearch, and the two python test applications
 # the test applications will use elasticcontributor's newfeature1 apm agent
-./script/compose.py start master --no-kibana --with-agent-python-django --with-agent-python-flask --python-agent-package=git+https://github.com/elasticcontributor/apm-agent-python.git@newfeature1 --force-build
+./scripts/compose.py start master --no-kibana --with-agent-python-django --with-agent-python-flask --python-agent-package=git+https://github.com/elasticcontributor/apm-agent-python.git@newfeature1 --force-build
```
beniwohli: good catch!

jahtalab:
Thanks @beniwohli! I've Made the requested changes!

jahtalab:
Failed tests:
* `test_conc_req_all_agents`
* `test_sidebar`

Seems unrelated to the changes!

tests/agent/test_rum.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def test_rum(rum):
+    elasticsearch = rum.apm_server.elasticsearch
+    elasticsearch.clean()
+    endpoint = Endpoint(rum.url, "run_integration_test")
+
+    r = requests.get(endpoint.url)
+    assert r.status_code == 200
+    utils.check_elasticsearch_transaction(elasticsearch, 1, {'query': {'term': {'processor.event': 'transaction'}}})
```
graphaelli: Is the number of spans generated here unpredictable?  If not, you could avoid having a custom query here.  I always get 7 docs locally.

tests/agent/test_rum.py
```diff
@@ -0,0 +1,13 @@
+from tests.endpoint import Endpoint
+from tests import utils
+import requests
+
+
+def test_rum(rum):
+    elasticsearch = rum.apm_server.elasticsearch
+    elasticsearch.clean()
+    endpoint = Endpoint(rum.url, "run_integration_test")
```
graphaelli: would it be possible to customize the response of that endpoint?  Perhaps with a query parameter?  We could use `check_agent_transaction` instead of duplicating it if this returned the text `run_integration_test` or something else static.

graphaelli:
this should pass following a rebase with master

tests/agent/test_rum.py
jahtalab: Unfortunately, the number of spans is not predictable. The reason is that we generate some of the spans based on the browser navigation timing and depending how fast or slow the test runs, the duration of some of the spans might be zero which we remove before generating the final payload.

tests/agent/test_rum.py
jahtalab: I'm not sure I understand! Currently the endpoint returns the request that was made to the apm-server (including the body) and the response that was received by the agent.

We can examine this response however since we already test agent <-> apm-server communication in the agent, I thought we shouldn't duplicate it here.

graphaelli: Sorry, that is indeed confusing.  This test duplicates`check_agent_transaction`, minus the `check_request_response` bit: https://github.com/elastic/apm-integration-testing/blob/46b8e044c95a638c34bc8a774e4b810b153ebe18/tests/utils.py#L23.  If the endpoint returned a predictable message, we could avoid that.

jahtalab: Ah, good point! I'll change the endpoint to accept a query parameter!

jahtalab:
Thanks for the review @graphaelli! I've made the changes!

PR #150, opened 2018-07-26
hardcode the healthcheck for opbeans-python and opbeans-node to 3000

beniwohli:
internal ports are all supposed to be 3000, so we can hardcode the
healthcheck as well


beniwohli:
failure is in `test_sidebar`, unrelated to this PR. Merging.

PR #151, opened 2018-07-26
first stab at creating a generic load generator

beniwohli:
The compose service gathers all services that need to have load generated
on (by default every opbeans service that is started) and exposes them
as a environment variable (`OPBEANS_URLS`) to the loadgen container.

In the entrypoint of the loadgen container, a `Procfile` is created on the
fly using `OPBEANS_URLS`, and then executed using honcho.

The loadgen container itself has been moved to its own [repo](https://github.com/elastic/opbeans-loadgen), and an automated Docker Hub build is set up.


docker/opbeans/workload/requirements.txt
```diff
@@ -0,0 +1,12 @@
+honcho==1.0.1
+molotov==1.6
+uvloop==0.11.0
+
```
graphaelli: does this space do something? just curious

scripts/compose.py
```diff
@@ -1168,6 +1168,46 @@ def _content(self):
...
+        # unless the no_opbeans_XYZ_flag is set
+        for flag, value in options.items():
+            if not value and flag.startswith('disable_opbeans_'):
+                service_name = flag[len('disable_'):]
+                if not options.get('no_{}_loadgen'.format(service_name)) and service_name != 'opbeans_load_generator':
+                    self.loadgen_services.append(service_name.replace('_', '-'))
+
+    @staticmethod
+    def enabled():
+        return True
```
graphaelli: I don't think this should be a default service - we want it to be enabled by default if any opbeans service is - that could be done he, like with postgres and redis: re https://github.com/elastic/apm-integration-testing/blob/cdd26cd2d4cb59eef5f80409bde0b45057d18219/scripts/compose.py#L2128-L2132

docker/opbeans/workload/requirements.txt
beniwohli: just trying to separate direct dependencies from their sub-dependencies

scripts/compose.py
beniwohli: to be honest, I'm not a huge fan how we (I think it was me in the original `local` script) solved this issue here. Depending on the existence of some flags, we add/remove some other flags at run-time. That seems wrong.

Not sure what a better solution would be. For the `--no` / `--with` flags, maybe we can simply always add them. Postgres/Redis and now load gen, I'm not so sure. The load gen container will immediately exit if there's no opbeans services running, so I don't think it would be terribly bad if it always started

graphaelli: I agree it's not the best solution but I think `argparse` puts us in a tough spot here (or I missed something).
The problem is that integration tests will start this and wait for all services to be healthy, and this won't be so they abort.  We either need to change this or update all of the wrapper scripts not to start it.

beniwohli:
@graphaelli do you mind giving this one another review treatment?

scripts/compose.py
```diff
@@ -1244,6 +1244,42 @@ def _content(self):
...
+                default=False,
+                help='Disable load generator for {}'.format(service_class.name())
+            )
+
+    def __init__(self, **options):
+        super(OpbeansLoadGenerator, self).__init__(**options)
+        self.loadgen_services = []
+        # create load for opbeans services
+        run_all_opbeans = options.get('run_all_opbeans')
+        excluded = ('opbeans_load_generator', 'opbeans_rum')
```
simitt: Could you add a note to the README about loadgen for RUM?

beniwohli:
@simitt thanks for the review, I added a note in the Readme. But to be honest, I don't think the script as it is should be recommended for usage outside of the APM team (I realize this is already happening). With the merger of integration testing, it became super complex and unwieldy (hundreds of flags, and about 3300 lines of code). A little bit of README that nobody reads won't help much.

I'm looking into splitting the whole thing up a bit again, hopefully with the result of having two scripts, one for local testing and one for integration testing, and shared code in a library.

beniwohli:
jenkins test this please

PR #152, opened 2018-07-26
use opbeans/opbeans-go base image for the opbeans-go service

beniwohli:
also, hardcode internal port to 3000

Unfortunately I had to hardcode port and cache URL in the `CMD`, since the image is based on the "Distroless", which lacks `/bin/sh`, which is needed for environment variable interpolation in `CMD`...


beniwohli:
test failure is unrelated to this change, merging

PR #153, opened 2018-07-27
use opbeans/opbeans-java base image, and hardcode internal port

beniwohli:


beniwohli:
build failure is unrelated, merging

PR #154, opened 2018-07-27
use opbeans/opbeans-ruby base image, and hardcode internal port to 3000

beniwohli:


scripts/compose.py
```diff
@@ -1912,9 +1913,9 @@ def test_opbeans_ruby(self):
                     volumes:
                       - .:/local-install
                     healthcheck:
-                      test: ["CMD", "curl", "--write-out", "'HTTP %{http_code}'", "--silent", "--output", "/dev/null", "http://opbeans-ruby:3001/"]
+                      test: ["CMD", "curl", "--write-out", "'HTTP %{http_code}'", "--silent", "--output", "/dev/null", "http://opbeans-ruby:3000/"]
                       interval: 5s
-                      retries: 12""")  # noqa: 501
+                      retries: 100""")  # noqa: 501
```
graphaelli: seems like a lot?

beniwohli: @graphaelli there's a comment at https://github.com/elastic/apm-integration-testing/pull/154/files#diff-d0df5dd5f0019c3b878fde30f4b7341bR1135 about this change. For some reason I'm not quite sure about, the app can take a few minutes to boot. I had failures even with 20 retries, so I bumped it to something safe

graphaelli: Ah, I didn't see the associated comment, sorry.  Fine with me since it's not in CI flow, I just thought it was debugging slipping in.

PR #157, opened 2018-07-28
add healthcheck to go integration test app

graphaelli:
non-issue so far but could possibly cause flakiness if tests are able to start running before the container is ready.


axw:
Is that output from inspecting the opbeans-go container? Or something else?
I just started up a stack with apm-integration-testing, using compose.py, and I get this:

```bash
docker inspect -f '{{.Config.Healthcheck | json}}' apm-integration-testing_opbeans-go
{"Test":["CMD","/opbeans-go","-healthcheck","localhost:3000"],"Interval":10000000000,"Timeout":3000000000,"Retries":10}
```

graphaelli:
This is the `gonethttpapp` not `opbeans-go` - I even confused myself in the last response

PR #160, opened 2018-08-03
add --build-parallel build option and use it on CI

graphaelli:
docker-compose build just got parallel build capabilities, let's try it out

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,8 @@ backports-abc==0.5
 certifi==2017.11.5
 chardet==3.0.4
 coverage==4.5.1
-docker-compose==1.21.1
+# for build --parallel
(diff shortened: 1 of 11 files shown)


requirements.txt
```diff
@@ -4,7 +4,7 @@ backports-abc==0.5
 certifi==2017.11.5
 chardet==3.0.4
 coverage==4.5.1
-docker-compose==1.21.1
+-e git://github.com/docker/compose@473703d0d9e0f81520dd41aa30da67d7868f877a#egg=docker-compose
```
axw: add a #comment that this is to get --build-parallel support?

graphaelli: great idea, done

PR #162, opened 2018-08-04
make dashboard loading depend on kibana

### LINKED ISSUE #161: make dashboard loading depend on kibana
If kibana is not started, disable/don't enable dashboard loading for apm-server, filebeat, and metricbeat.  Right now [in apm-integration-testing] apm-server doesn't load dashboards by default because of this dependency handling.  Likewise, filebeat and metricbeat will not start without kibana since they are always set to load dashboards.

graphaelli:
* load apm-server dashboards by default for now, probably to default to false for >= 6.5

closes #161

### DIFF
--- a/scripts/compose.py
+++ b/scripts/compose.py
@@ -321,11 +321,10 @@ class StackService(object):
                 help="stack {} override".format(image_detail_key),
             )
 
+
 #
 # Elastic Services
 #
-
-
 class ApmServer(StackService, Service):
     docker_path = "apm"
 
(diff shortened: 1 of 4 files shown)


scripts/compose.py
```diff
@@ -388,16 +395,23 @@ def add_arguments(cls, parser):
             default='elasticsearch',
             help='apm-server output'
         )
+        parser.add_argument(
+            "--no-apm-server-dashboards",
+            action="store_false",
+            dest="apm_server_dashboards",
+            help="apm-server output",
```
jalvz: Maybe some more useful help text, here and/or in readme, also mentioning that it doesn't take effect kibana is disabled
Particularly, if we are pointing non apm devs to this repo...

graphaelli: nice catch, fixing

PR #164, opened 2018-08-13
refactor how --with-/--no- flags work a bit

beniwohli:
instead of changing which flags are available at run time, we
put them into mutually exclusive groups

We also invert the target flag name from `disable_X` to `enable_X`,
which gets rid of a whole lot of double negatives.

One drawback of this change is that something like

    compose.py start master --all --no-opbeans-python

doesn't start all but the opbeans-python services anymore. `--all` trumps
all now.

### DIFF
--- a/scripts/compose.py
+++ b/scripts/compose.py
@@ -354,7 +354,7 @@ class ApmServer(StackService, Service):
         ]
         self.depends_on = {"elasticsearch": {"condition": "service_healthy"}}
 
-        if not self.options.get("disable_kibana", False):
+        if self.options.get("enable_kibana", True):
             self.depends_on["kibana"] = {"condition": "service_healthy"}
             if options.get("apm_server_dashboards", True):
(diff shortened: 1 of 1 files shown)


scripts/compose.py
```diff
@@ -2464,8 +2460,15 @@ def start_handler(self):
             args["version"] = self.SUPPORTED_VERSIONS.get(args["stack-version"], args["stack-version"])
 
         selections = set()
+        all_opbeans = args.get('run_all_opbeans')
+        any_opbeans = all_opbeans or any(v and k.startswith('enable_opbeans_') for k, v in args.items())
         for service in self.services:
-            if not args.get("disable_" + service.option_name()):
+            if args.get("enable_" + service.option_name()) \
```
jalvz: you could trow here some variable to make the whole thing more readable...

beniwohli: excellent point. Done in 166fce39

PR #166, opened 2018-08-16
only use our homegrown downloader for BC images [WIP]

beniwohli:
images of snapshot builds are now pushed to docker.elastic.co, so we can
use that.

Also, only build services that are actually used in this run.

6.4 and 6.3 snapshots haven't been pushed so far, so waiting to merge this

### DIFF
--- a/scripts/compose.py
+++ b/scripts/compose.py
@@ -265,21 +265,15 @@ class StackService(object):
 
     def image_download_url(self):
         # Elastic releases are public
-        if self.release:
+        if not self.bc:
             return
 
         version = self.version
         image = self.docker_name
         if self.oss:
             image += "-oss"
-        if self.bc:
-            return "https://staging.elastic.co/{version}-{sha}/docker/{image}-{version}.tar.gz".format(
-                sha=self.bc,
-                image=image,
-                version=version,
-            )
-
-        return "https://snapshots.elastic.co/docker/{image}-{version}-SNAPSHOT.tar.gz".format(
+        return "https://staging.elastic.co/{version}-{sha}/docker/{image}-{version}.tar.gz".format(
+            sha=self.bc,
             image=image,
             version=version,
         )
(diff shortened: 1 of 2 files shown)


scripts/compose.py
```diff
@@ -1676,14 +1670,18 @@ def start_handler(self):
...
                 if args["force_build"]:
                     docker_compose_build.append("--no-cache")
                 if args["build_parallel"]:
                     docker_compose_build.append("--parallel")
-                subprocess.call(docker_compose_build)
+                subprocess.call(docker_compose_build + build_services)
 
+            # pull any images
+            image_services = [name for name, service in compose["services"].items() if 'image' in service]
+            subprocess.call(["docker-compose", "-f", docker_compose_path.name, "pull"] + image_services)
```
graphaelli: Why separate the build and pull commands?  Won't build without specifying the list of services pull anything needed?

scripts/compose.py
```diff
@@ -265,21 +265,15 @@ class StackService(object):
 
     def image_download_url(self):
         # Elastic releases are public
-        if self.release:
+        if not self.bc:
```
graphaelli: If you want to get this merged soon we could add a version limit here too, eg `self.at_least_version('6.5')`

scripts/compose.py
beniwohli: `docker-compose build` doesn't do anything for image-only services. And `docker-compose up` doesn't have a `--pull` flag (https://github.com/docker/compose/issues/3574 if you want to laugh/cry/despair). So AFAIK, separating it up is the only way we can do this

graphaelli: Aha!  And this is needed because previously we didn't have images from a registry that would change but now we do - this is great then, thanks for explanation.

scripts/compose.py
beniwohli: 6.3.3 and 6.4 snapshots are now also available in the registry. I guess that should be fine? For 6.2 we use `--release` anyway, right?

graphaelli: That's great news.  Agreed this should be fine as-is.

PR #167, opened 2018-08-23
add --apm-server-build option

graphaelli:
As a developer, I'd like to run apm-server still in development under docker.  That enables other services, like opbeans and integration test apps, that typically run under docker, to easily communicate with that development version of apm-server.

This change introduces a `--apm-server-build` option to `compose.py` that will build an apm-server container image that is close enough to a [real](apm-server-docker) container image for testing.

## Example usage

```
./scripts/compose.py start 6.4 --release --apm-server-build=https://github.com/elastic/apm-server.git@v2
```
builds `apm-integration-testing_apm-server` from `docker.elastic.co/apm/apm-server:6.4.0`

```
$ ./scripts/compose.py versions
Getting current version numbers for services...

APM Server (image built: 2018-08-23 22:53:20 UTC):
	apm-server version 7.0.0-alpha1 (amd64), libbeat 7.0.0-alpha1 [88e1e0ebf6c6c020bc02149ec79f4a1846904653 built 2018-08-23 22:53:02 +0000 UTC]

Elasticsearch (image built: 2018-08-18 00:13:27 UTC):
	OpenJDK 64-Bit Server VM warning: Option UseConcMarkSweepGC was deprecated in version 9.0 and will likely be removed in a future release.
Version: 6.4.0, Build: default/tar/595516e/2018-08-17T23:18:47.308994Z, JVM: 10.0.2

Kibana (image built: 2018-08-18 00:20:48 UTC):
	Version: 6.4.0
	Branch: 6.x
	Build SHA: 07559c1e1053d20fe543a4845a2311bb9916cd2c
	Build number: 17929
```

PR #168, opened 2018-08-24
add --kibana-with-canvas option

graphaelli:
#167 demonstrated a straightforward path to customizing container images.  Here's one for adding canvas to kibana.

## Example usage

```
./scripts/compose.py start 6.4 --release --kibana-with-canvas
```

builds `apm-integration-testing_kibana` from `docker.elastic.co/kibana/kibana:6.4.0`

`--kibana-with-canvas` takes an optional argument to use a different build version.

### DIFF
--- a/scripts/compose.py
+++ b/scripts/compose.py
@@ -526,11 +526,15 @@ class Kibana(StackService, Service):
     @classmethod
     def add_arguments(cls, parser):
         super(Kibana, cls).add_arguments(parser)
+        canvas_default_version = "0.1.2174"
         parser.add_argument(
             "--kibana-with-canvas",
-            const="2174",
+            const=canvas_default_version,
             nargs="?",
-            help="build custom kibana image with canvas",
+            help=(
+                "build custom kibana image with canvas, "
+                "override default version {!r} by providing an argument.".format(canvas_default_version)
+            ),
         )
 
     def _content(self):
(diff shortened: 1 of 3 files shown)


scripts/compose.py
```diff
@@ -522,13 +523,35 @@ def __init__(self, **options):
             if self.at_least_version("6.3"):
                 self.environment["XPACK_XPACK_MAIN_TELEMETRY_ENABLED"] = "false"
 
+    @classmethod
+    def add_arguments(cls, parser):
+        super(Kibana, cls).add_arguments(parser)
+        parser.add_argument(
+            "--kibana-with-canvas",
+            const="2174",
```
beniwohli: can you add a comment here what this number signifies? I assume it is the patch-level version number

Also, looking at the interplay between `nargs="?"` and `const`, it will `--kibana-with-canvas` will result in `0.1.2174` being build, while `--kibana-with-canvas 1234` will build `0.1.1234`, right? Might make sense to explain that in `help`

graphaelli: Sure, I'll expand the help.  I was just messing around with this over lunch and didn't expect anyone to actually look at it :) (thank you)

graphaelli: Updated help and switched to accepting the entire version.

PR #170, opened 2018-08-31
added flags to loadgen to set specific RPM per service

beniwohli:
this depends on elastic/opbeans-loadgen#3


beniwohli:
jenkins test this please

PR #172, opened 2018-09-11
Revert opbeans node to use old loadgen

watson:


watson:
I'm not entirely sure why CI is failing. I can't see anything obvious in the logs.

PR #178, opened 2018-09-12
use /api/status for kibana healthcheck, and use --fail for the curl cmd

beniwohli:
This ensures that the health check will fail on a non-200 status code


graphaelli:
Of course, needs a minor update to the tests

PR #179, opened 2018-09-12
Jrs/quickstart

jamiesmith:
This PR takes the relevant start/stop info from the README and distills it down to a subset suitable for folks just wanting to play with or demonstrate the env, rather than development and test.

### DIFF
--- a/QUICKSTART.md
+++ b/QUICKSTART.md
@@ -1,0 +1,133 @@
+# APM LocalEnv Quickstart
+
+In addition to the end to end (eg agent -> apm server -> elasticsearch <- kibana) development and testing of Elastic APM, this repo also serves as a nice way to spin up a local environment of Elastic APM.  The [README](/README.md) has very detailed instructions, but focuses mostly on using it for development.  This doc really just concentrates on getting it running.  For advanced topics, check out the main README.
+
+Note that by "local environment", this can be on your actual local machine, or running on a cloud instance with port forwards set up.
+
+[![Build Status](https://apm-ci.elastic.co/view/All/job/elastic+apm-integration-testing+master+push/badge/icon?style=plastic)](https://apm-ci.elastic.co/job/elastic+apm-integration-testing+master+push/)
+
+## Prerequisites
+
+The basic requirements for starting a local environment are:
+
+- Docker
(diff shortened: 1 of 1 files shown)


QUICKSTART.md
```diff
@@ -0,0 +1,131 @@
...
+## Running Local Enviroments
+
+### Starting an Environment
+
+The tool that we use to start and stop the environment is `scripts/compose.py`.  This provides a handy cli for starting an APM environment using docker-compose. 
+
+#### TL;DR
+
+Start an env by running:
+`./scripts/compose.py start --all 6.4`
```
graphaelli: Perhaps recommend `--release` for quickstart

QUICKSTART.md
```diff
@@ -0,0 +1,131 @@
...
+  --with-opbeans-java   Enable opbeans-java
+  --with-opbeans-node   Enable opbeans-node
+  --with-opbeans-python
+  --with-opbeans-ruby   Enable opbeans-ruby
+  --with-opbeans-rum    Enable opbeans-rum
+```
+So when new agents are added we don't have to update these instructions.
+
+
+**Bonus**:  With either the `all` or individual methods above, you can also pass a `--with-metricbeat` flag, which will also set up metricbeat.  One side note here is that you will need to then probably need to set a default index pattern
```
graphaelli: `--with-filebeat` might also be worth a mention

QUICKSTART.md
```diff
@@ -0,0 +1,131 @@
...
+
+#### Status
+
+Each app gets its own port.  You can actually hit them with your browser.  They all have a similar look & feel.
+
+You can check the status of your APM cluster with `/scripts/compose.py status`, which basically calls :
```
graphaelli: either drop the `/` or prefix with `./` - usage is mixed throughout this doc, would be good to sync them all up.

QUICKSTART.md
jamiesmith: done, and fixed the wording around it

QUICKSTART.md
jamiesmith: Will do.  That one was a copy paste.  Nice catch

PR #182, opened 2018-09-18
opbeans: allow new customers to be insert

watson:
The `customer_pk` counter was set to start at 1 even though there was already a 1000 customers in the database. This meant that if you tried to insert a new customer you would get a duplicate key error.

This commit also leaves the sql uncompressed so in the future it's possible to see the changes made in a diff


kuisathaverat:
@watson the opbeans.sql or opbeans.sql.gz file is not used in any place in the integration test, I did not found any reference on files. I saw that the opbean-node project has another schema files probably more updated than this one https://github.com/elastic/opbeans-node/tree/master/db 

Is this PR still useful?

watson:
Good question. I can't remember honestly. I don't think the opbeans-node schema files are any newer than this one though.

Though there's no reference to the actual file, there is a reference to the directory: https://github.com/elastic/apm-integration-testing/blob/822720ffc4c0089b3d679055427b6fc6c5487761/scripts/compose.py#L868

I'm not entirely sure what that section of the compose.py file means though. Could that somehow be using the file inside the directory?

kuisathaverat:
actually, it seems used, so let's merge it

```
Initialization scripts
If you would like to do additional initialization in an image derived from this one, add one or more *.sql, *.sql.gz, or *.sh scripts under /docker-entrypoint-initdb.d (creating the directory if necessary). After the entrypoint calls initdb to create the default postgres user and database, it will run any *.sql files, run any executable *.sh scripts, and source any non-executable *.sh scripts found in that directory to do further initialization before starting the service.
```

PR #184, opened 2018-10-01
Remove memory limits for ES container

sqren:
Why does elasticsearch, as the only container, have memory restriction on the Docker level?
Also, can we increase the heap size to 5gb?

### DIFF
--- a/scripts/compose.py
+++ b/scripts/compose.py
@@ -468,8 +468,8 @@ class ApmServer(StackService, Service):
 class Elasticsearch(StackService, Service):
     default_environment = ["cluster.name=docker-cluster", "bootstrap.memory_lock=true", "discovery.type=single-node"]
     default_es_java_opts = {
-        "-Xms": "1g",
-        "-Xmx": "1g",
+        "-Xms": "5g",
+        "-Xmx": "5g",
     }
 
     SERVICE_PORT = 9200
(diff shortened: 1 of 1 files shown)


scripts/compose.py
```diff
@@ -503,7 +503,6 @@ def _content(self):
                 "retries": 10,
                 "test": ["CMD-SHELL", "curl -s http://localhost:9200/_cluster/health | grep -vq '\"status\":\"red\"'"],
             },
-            mem_limit="5g",
```
graphaelli: agreed on this

scripts/compose.py
```diff
@@ -468,8 +468,8 @@ def render_proxy(self):
 class Elasticsearch(StackService, Service):
     default_environment = ["cluster.name=docker-cluster", "bootstrap.memory_lock=true", "discovery.type=single-node"]
     default_es_java_opts = {
-        "-Xms": "1g",
-        "-Xmx": "1g",
+        "-Xms": "5g",
+        "-Xmx": "5g",
```
graphaelli: Won't this be too large of a default for most people?  Default Docker for mac setting is 2 GB memory + 1 GB swap, I'm not sure if these settings will cause an immediate problem though.

I can commit to implementing https://github.com/elastic/apm-integration-testing/blob/9497bb2a70a9c449219b207e7126e5277f200ae0/scripts/compose.py#L483 this week, in case that affects your response.

sqren: I'm currently modifying the compose.py file, so having command line options for this would be great!

I'll leave the default as-is.

PR #185, opened 2018-10-02
add --elasticsearch-heap and --elasticsearch-java-opts args

graphaelli:
follow up on #184


graphaelli:
fyi @sqren `--elasticsearch-heap 5g` should do the trick now

sqren:
Thanks!

PR #186, opened 2018-10-02
add --apm-server-self-instrument option

graphaelli:


graphaelli:
Per discussion, switched to enable by default and added `--no-apm-server-self-instrument`.  The test failure is specific to some java agent changes and unrelated to this pr.

PR #195, opened 2018-10-17
docker/go/nethttp: update to go.elastic.co/apm

axw:
Update the test program to use the new go.elastic.co/apm package import path and name.


axw:
Not to be merged until the apm-agent-go change is merged into master.

axw:
The test failures are at least in part due to the transactions that apm-server sends itself being counted. I'll fix that in another PR.

Issue #196, opened 2018-10-17
Opbeans: implement randomized distributed tracing


beniwohli:
To have somewhat realistic test data for the UI, the Opbeans services should implement distributed tracing amongst each other. The plan is as follows:

 * when using `compose.py`, an environment variable, `OPBEANS_SERVICES` will be set for all opbeans services, containing a list of all opbeans services in a comma-separated list, e.g. 
    
        OPBEANS_SERVICES=opbeans-node,opbeans-python,opbeans-go
    
   these services can either be:
     * "opbeans-xyz", in which case the base URL `http://opbeans-xyz:3000` should be used
     * or an URL that starts with `http`, in which case the URL should be used verbatim as base URL
       
* when a GET request to `/api/*` comes in, the opbeans service flips a coin. Depending on the outcome, it either serves the request itself, or forwards the request via instrumented HTTP-request to one of the services from `OPBEANS_SERVICES` using the URL `http://opbeans-XYZ:3000/original/url`.

* there will be a parameter to define the probability of relaying the call to another node (DT). The default value will be 0.5 and 0 will disable it.

      OPBEANS_DT_PROBABILITY=0.5

**Note**: some opbeans implementations have non-standard endpoints in the `/api/` namespace. These shouldn't be affected by the randomized forwarding

Implemented in:

 * [x] `compose.py` (https://github.com/elastic/apm-integration-testing/pull/197)
 * [x] opbeans-node (https://github.com/elastic/opbeans-node/pull/40)
 * [x] opbeans-ruby (https://github.com/elastic/opbeans-ruby/pull/8)
 * [x] opbeans-java (https://github.com/elastic/opbeans-java/pull/14)
 * [x] opbeans-go (https://github.com/elastic/opbeans-go/pull/10)
 * [x] opbeans-python (https://github.com/elastic/opbeans-python/pull/18)

mikker:
For good measure: What are the _standard_ endpoints in `/api`?

watson:
For reference, here's how we have split up `/api/*` into its own service for opbeans-node: https://github.com/elastic/opbeans-node/pull/36

alvarolobato:
What percentage should be used when flipping a coin? 50%?

watson:
The opbeans-node agent have the following HTTP endpoints:

- `GET /is-it-coffee-time` - Not called by the UI and probably unique to opbeans-node
- `GET /log-error` - Not called by the UI and probably unique to opbeans-node
[...]

PR #197, opened 2018-10-17
added OPBEANS_SERVICES environment variable

beniwohli:
refs #196


scripts/compose.py
```diff
@@ -1758,6 +1758,13 @@ def start_handler(self):
         services = {}
         for service in selections:
             services.update(service.render())
+
+        # expose a list of enabled opbeans services to all opbeans services. This allows them to talk amongst each other
+        # and have a jolly good distributed time
+        enabled_opbeans_services = [k for k in services.keys() if k.startswith("opbeans-") and not k in ("opbeans-rum", "opbeans-load-generator")]
+        for s in enabled_opbeans_services:
+            services[s]["environment"].append("OPBEANS_SERVICES=" + ",".join(enabled_opbeans_services))
```
graphaelli: environment can also be a `dict` - not sure if any opbeans services have that, yet

beniwohli: fixed in 4e4f3c5

scripts/compose.py
```diff
@@ -1758,6 +1758,17 @@ def start_handler(self):
         services = {}
         for service in selections:
             services.update(service.render())
+
+        # expose a list of enabled opbeans services to all opbeans services. This allows them to talk amongst each other
+        # and have a jolly good distributed time
+        enabled_opbeans_services = [k for k in services.keys() if k.startswith("opbeans-") and not k in ("opbeans-rum", "opbeans-load-generator")]
```
kuisathaverat: maybe we want only to enable DT on some test, we could add a parameter for that something like this

if options.get('opbeans_dt_enable'):

beniwohli: how about we kill two birds with one stone* and use a `--opbeans-dt-probability` float flag? Setting it to 0 disables dt, default will be 0.5

* no birds have been killed during the production of this comment

beniwohli: @kuisathaverat implemented in 4d1df35

scripts/compose.py
```diff
@@ -1758,6 +1772,18 @@ def start_handler(self):
         services = {}
         for service in selections:
             services.update(service.render())
+
+        # expose a list of enabled opbeans services to all opbeans services. This allows them to talk amongst each other
+        # and have a jolly good distributed time
+        enabled_opbeans_services = [k for k in services.keys()
+                                    if k.startswith("opbeans-") and k not in ("opbeans-rum", "opbeans-load-generator")]
```
alvarolobato: @beniwohli does this include the ports? I'm testing with opbeans java and I'm not getting the ports.

beniwohli:
Test failure  is unrelated, merging

PR #198, opened 2018-10-18
unify handling of repo/branch/local flags

beniwohli:
This adds `--opbeans-XYZ-agent-branch`, `--opbeans-XYZ-agent-repo` and `--opbeans-XYZ-agent-local-repo` to all backend Opbeans, and removes the partially overlapping custom flags

This is somewhat backwards incompatible, as it removes default branches/repos from the environment if they are not explicitly set. The idea is that in this case the opbeans containers should simply install the newest version from their respective package managers.


scripts/compose.py
```diff
@@ -1178,16 +1187,6 @@ def add_arguments(cls, parser):
             default=cls.DEFAULT_LOCAL_REPO,
         )
 
-    def __init__(self, **options):
-        super(OpbeansPython, self).__init__(**options)
-        self.local_repo = options.get("opbeans_python_local_repo", self.DEFAULT_LOCAL_REPO)
```
beniwohli: I removed this special handling, assuming that nobody really needs to test 6.2 with opbeans anymore. If they do, they can use the branch flag manually

beniwohli:
jenkins test this

PR #203, opened 2018-10-19
test agent error reporting

### LINKED ISSUE #30: Add Tests for Errors
Currently only transactions are sent and checked

graphaelli:
Documents current integration testing application API and adds a new endpoint, `/oof`, that should trigger auto-instrumented error capturing.

closes #30

### DIFF
--- a/docker/nodejs/express/app.js
+++ b/docker/nodejs/express/app.js
@@ -48,13 +48,13 @@ function extra_route () {
     span.end()
 }
 
+app.get("/oof", function(req, res, next) {
+    next(new Error("oof"));
+});
+
 app.use(function (err, req, res, next) {
   console.error(err.stack)
   res.status(500).send('Something broke!')
-});
-
-app.get("/oof", function(req, res, next) {
-    next(new Error("oof"));
 });
 
 var server = app.listen(process.env.EXPRESS_PORT, '0.0.0.0', function () {
--- a/docker/nodejs/express/app.js
+++ b/docker/nodejs/express/app.js
@@ -48,6 +48,15 @@ function extra_route () {
     span.end()
 }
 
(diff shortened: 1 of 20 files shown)


docker/nodejs/express/app.js
```diff
@@ -48,6 +48,15 @@ function extra_route () {
     span.end()
 }
 
+app.use(function (err, req, res, next) {
+  console.error(err.stack)
+  res.status(500).send('Something broke!')
+});
```
Qard: Put this after the `/oof` route and it should work.

tests/agent/test_python.py
```diff
@@ -7,8 +7,13 @@
...
 def test_req_flask(flask):
-    utils.check_agent_transaction(
-        flask.foo, flask.apm_server.elasticsearch, ct=2)
+    utils.check_agent_transaction(flask.foo, flask.apm_server.elasticsearch)
+
+
+@pytest.mark.version
+@pytest.mark.flask
+def test_flask_error(flask):
+    utils.check_agent_error(flask.oof, flask.apm_server.elasticsearch, ct=2)
```
graphaelli: @beniwohli I was surprised this generated two documents

beniwohli: I haven't looked at the actual documents, but this is probably due to [`logging=True`](https://github.com/elastic/apm-integration-testing/pull/203/files#diff-b4c662e3253b95a7af5361efeee2512dR24), which automatically sets up logging using our `LoggingHandler`. So one error is from our exception handler, and the other one from the logging handler, at least if my guess is right.

graphaelli: Looks right based on the documents produced, good with that for the tests?

PR #220, opened 2018-11-06
agent integration tests with remote apm-server

graphaelli:
This change enables agents to communicate with a remote, secret token enabled apm-server, primarily for use with cloud integration tests.

Follows on from #219


graphaelli:
jenkins, retest this please

docker/nodejs/express/app.js
```diff
@@ -4,8 +4,6 @@ var apm = require('elastic-apm-node').start({
   serviceName: process.env.EXPRESS_SERVICE_NAME,
   flushInterval: 1,
   maxQueueSize: 1,
-  secretToken: '1234',
```
simitt: why did you remove the `secretToken` here instead of also reading it from the env var?

graphaelli: Since there is nothing specific to integrating with express here, I moved to the simpler environment variable configuration for the nodejs agent.

This is in contrast with the flask and django integrations, where those settings are framework specific and expected to be ultimately configure the python agent.

I think I got all of the `secret_token` configs, eg the django one - https://github.com/elastic/apm-integration-testing/pull/220/files#diff-ea2db610473460b85666c49665c7893fR47 - if you could audit I'd appreciate it.

PR #229, opened 2018-11-14
Add route change load generation to opbeans-rum

jahtalab:
Based on: https://github.com/elastic/opbeans-frontend/commit/7962b30a762f36c6b9c21b66970087ce9fa03035

### DIFF
--- a/docker/opbeans/rum/tasks.js
+++ b/docker/opbeans/rum/tasks.js
@@ -3,32 +3,38 @@ const { Chromeless } = require('chromeless')
 var baseUrl = process.env.OPBEANS_BASE_URL || 'http://www.opbeans.com';
 var url = baseUrl
 
-function sleep(ms){
-    return new Promise(resolve=>{
-        setTimeout(resolve,ms)
+function sleep(ms) {
+    return new Promise(resolve => {
+        setTimeout(resolve, ms)
     })
 }
 
 async function run() {
-  const chromeless = new Chromeless({
-      lauchChrome: true
-  })
+    const chromeless = new Chromeless({
+        lauchChrome: true
+    })
 
-  for(;;){
-    url = await chromeless
-    .goto(url)
-    .evaluate((baseUrl) => {
-        var links = document.querySelectorAll('a[href^="/"]')
-        if (links && links.length) {
-            var i = Math.floor(Math.random()*links.length)
-            return links[i].href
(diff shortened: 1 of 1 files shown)


docker/opbeans/rum/tasks.js
```diff
@@ -3,32 +3,37 @@ const { Chromeless } = require('chromeless')
...
-  })
+    const chromeless = new Chromeless({
+        lauchChrome: true
+    })
+
+    for (; ;) {
+        url = await chromeless
+            .goto(url)
+            .evaluate((baseUrl) => {
+                if (typeof window.activateLoadGeneration === 'function') {
```
beniwohli: maybe add a comment here that points the clueless reader (aka me) to https://github.com/elastic/opbeans-frontend/blob/7962b30a762f36c6b9c21b66970087ce9fa03035/src/rum.js#L28

jahtalab: @beniwohli , I'm not sure about adding a reference to a specific line since it will be outdated quickly.

jahtalab: But I think I should've provided that link in this PR's comment, I've updated the description

beniwohli: @jahtalab didn't mean to make a comment to a specific line, more like "this function is defined in opbeans-frontend" or some such. I was stumped for a while because I couldn't find the function in this repo

jahtalab: I thought since it's on `window` and inside the `chromless.evaluate` it should be clear it's in the frontend. But you can always be more clear 😄 . I've added the comment.

Thanks for the review!

jahtalab:
Test failure seems to be related to the go agent, 

`Service 'agent-go-net-http' failed to build: The command '/bin/sh -c go get github.com/elastic/apm-agent-go' returned a non-zero code: 1`

axw:
@jahtalab because of https://github.com/elastic/apm-integration-testing/pull/195, I'm working on it now. It seems that there are some other issues too, like the transactions the apm-server sends itself being included in counts.

axw:
I've merged #195, and just opened https://github.com/elastic/apm-integration-testing/pull/230 to fix the remaining issue.

jahtalab:
The test failures are unrelated to the change!

PR #230, opened 2018-11-15
tests: exclude apm-server events from counts

axw:
When running tests, ignore the tracing events that the apm-server sends itself by default. They are being incorrectly counted in the agent trace events.


axw:
I think my prior analysis was wrong. CI passes `--no-apm-server-self-instrument` to disable APM server instrumentation.

PR #231, opened 2018-11-16
updated python test matrix to only test latest release with 6.5+

beniwohli:
Version 4 of the agent doesn't support earlier APM Server versions

/edit also added excludes for NodeJS and Ruby. @watson @mikker please review :)


beniwohli:
looks like the build is stuck in a timeout :(

beniwohli:
jenkins test this please

PR #234, opened 2018-11-20
use apt to install chrome in Dockerfile

beniwohli:
this solves a dependency issue that dpkg on its own didn't resolve


beniwohli:
merging, since (upcoming) failure is unrelated. The build of the image itself worked.

PR #235, opened 2018-11-20
buildSpan is no more

beniwohli:
also, set api_request_time to a low value to avoid timeouts

### DIFF
--- a/docker/nodejs/express/app.js
+++ b/docker/nodejs/express/app.js
@@ -4,6 +4,7 @@ var apm = require('elastic-apm-node').start({
   serviceName: process.env.EXPRESS_SERVICE_NAME,
   flushInterval: 1,
   maxQueueSize: 1,
+  api_request_time: "50ms",
   ignoreUrls: ['/healthcheck']
 })
 
@@ -23,8 +24,7 @@ app.get("/foo", function(req, res) {
 });
 
 function foo_route () {
-    var span = apm.buildSpan()
-    span.start('app.foo')
+    var span = apm.startSpan('app.foo', 'custom')
     span.end()
 }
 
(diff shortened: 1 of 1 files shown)


docker/nodejs/express/app.js
```diff
@@ -4,6 +4,7 @@ var apm = require('elastic-apm-node').start({
   serviceName: process.env.EXPRESS_SERVICE_NAME,
   flushInterval: 1,
   maxQueueSize: 1,
+  api_request_time: "50ms",
```
watson:
```suggestion
  apiRequestTime: "50ms",
```

docker/nodejs/express/app.js
```diff
@@ -23,8 +24,7 @@ app.get("/foo", function(req, res) {
 });
 
 function foo_route () {
-    var span = apm.buildSpan()
-    span.start('app.foo')
+    var span = apm.startSpan('app.foo', 'custom')
```
watson: `custom` type is the default, so no need to set it 😃 

```suggestion
    var span = apm.startSpan('app.foo')
```

docker/nodejs/express/app.js
```diff
@@ -34,15 +34,13 @@ app.get("/bar", function(req, res) {
 });
 
 function bar_route () {
-    var span = apm.buildSpan()
-    span.start('app.bar')
+    var span = apm.startSpan('app.bar', 'custom')
```
watson:
```suggestion
    var span = apm.startSpan('app.bar')
```

docker/nodejs/express/app.js
```diff
@@ -34,15 +34,13 @@ app.get("/bar", function(req, res) {
...
     extra_route()
     span.end()
 }
 
 function extra_route () {
-    var span = apm.buildSpan()
-    span.start('app.extra')
+    var span = apm.startSpan('app.extra', 'custom')
```
watson:
```suggestion
    var span = apm.startSpan('app.extra')
```

docker/nodejs/express/app.js
beniwohli: oops

docker/nodejs/express/app.js
beniwohli: I was going with "let's be explicit in tests". OK to leave it in?

watson: Up to you 👍

beniwohli:
didn't fix the tests yet, but cut down the console output from 15mbyte to a few hundred kilobyte, so I'll chalk that up as a success

PR #236, opened 2018-11-20
Limit api_request_time to 50ms for all integration test apps

beniwohli:
For nodejs, this is done in #235


mikker:
@beniwohli Ok – we can leave it there then 👍

PR #239, opened 2018-11-21
give events more time to reach elasticsearch

graphaelli:


graphaelli:
#236 makes this unnecessary.

PR #241, opened 2018-11-21
set nodejs/express test agent framework explicitly

graphaelli:
Test whether this resolves test failures.  Probably should not be merged as it indicates framework detection might be actually be broken.

### DIFF
--- a/docker/nodejs/express/app.js
+++ b/docker/nodejs/express/app.js
@@ -1,6 +1,8 @@
 'use strict'
 
 var apm = require('elastic-apm-node').start({
+  frameworkName: 'express',
+  frameworkVersion: 'unknown',
   serviceName: process.env.EXPRESS_SERVICE_NAME,
   flushInterval: 1,
   maxQueueSize: 1,


beniwohli:
@watson without it, the framework isn't set in the integration tests, which would indeed indicate some kind of regression.

watson:
I don't have time to fix this in the agent right now. I've created [an issue](https://github.com/elastic/apm-agent-nodejs/issues/688) for now. If this is a blocking, feel free to merge for now and I'll revert it once it's fixed in the agent.

Btw, is this a bug with both major versions of the agent, or only 1.x or 2.x?

simitt:
With https://github.com/elastic/apm-server/pull/1546 the `framework.name` and `framework.version` are not required anymore. Maybe we should remove the check for the `framework` from the integration test then? This way we would still test that the agent sends up a `framework.name` and `version` where it was required (as otherwise the events would be rejected).

graphaelli:
Since these tests are all using frameworks I think we should keep the check in there.  I do think we should introduce a way to mark the individual test to skip this check instead of the [blanket exclude](https://github.com/elastic/apm-integration-testing/blob/c8866b20f7cf09b8e92a3d5ae8f91eef1b831f10/tests/agent/concurrent_requests.py#L198) at the agent level done now.  

I'm happy to continue that discussion but would like to merge this in now to get the tests passing again.

PR #244, opened 2018-11-27
add workaround to opbeans-go to ensure docker layer invalidation

beniwohli:
this fetches a JSON file from the GitHub API, the content of which
changes if the branch changes. With this trick, we can ensure that local
docker layer caches are invalidated when the remote branch is updated.

A drawback of this is that it will stop us from building the image if
the GitHub API is not accessible.


beniwohli:
@kuisathaverat can't claim I came up with it unfortunately :D https://stackoverflow.com/a/39278224/45691

axw:
Thanks @beniwohli!

PR #252, opened 2018-12-19
use new healthcheck with apm-server >= 6.5

graphaelli:
to eliminate the `/healthcheck' endpoint is deprecated, please use '/' instead` logging


graphaelli:
jenkins, retest this please

PR #254, opened 2018-12-20
discover latest build candidate

graphaelli:
It's a hassle to look this up manually.

### DIFF
--- a/scripts/compose.py
+++ b/scripts/compose.py
@@ -125,13 +125,14 @@ def curl_healthcheck(port, host="localhost", path="/healthcheck",
 def resolve_bc(version, kind):
     if kind != "latest":
         return kind
-    logging.debug("searching for latest build candidate for %s", version)
     branch = ".".join(version.split(".", 2)[:2])
     rsp = urlopen("https://staging.elastic.co/latest/{}.json".format(branch))
     if rsp.code != 200:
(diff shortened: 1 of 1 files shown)


scripts/compose.py
```diff
@@ -121,6 +122,23 @@ def curl_healthcheck(port, host="localhost", path="/healthcheck",
...
+    logging.debug("searching for latest build candidate for %s", version)
+    branch = ".".join(version.split(".", 2)[:2])
+    rsp = urlopen("https://staging.elastic.co/latest/{}.json".format(branch))
+    if rsp.code != 200:
+        raise Exception("failed to query build candidates at {}: {}".format(rsp.geturl(), rsp.info()))
+    encoding = "utf-8" # python2 rsp.headers.get_content_charset("utf-8")
+    info = json.load(codecs.getreader(encoding)(rsp))
+    try:
+        build_id = info["build_id"]
+        return build_id[build_id.rfind("-")+1:]
```
beniwohli: It might be helpful to print something here to tell the user that we found a BC and are downloading it. Unfortunately the JSON doesn't contain a whole lot of information (build time/date would be cool), but we could print the summary.html URL for the BC that we selected.

graphaelli: 👍 will add it before merging

PR #255, opened 2018-12-25
generate 6.x/7.x compatible search response

graphaelli:
The original fix in #249 only works against 7.x.  It throws this exception against 6.x:
```
08:46:01 >           if "value" in total:
08:46:01 E           TypeError: argument of type 'int' is not iterable
```

reverts 949a24b5ef4b0fa7b6a85a1eb989c2a769b74bd6


graphaelli:
turns out this doesn't work with all of es 6.x, returning a 400 Bad Request:

```
illegal_argument_exception', 'request [/apm-*/_search] contains unrecognized parameter: [rest_total_hits_as_int]')
```

will revert and improve the handling

Issue #261, opened 2019-01-07
support release differences more sustainably


graphaelli:
This repo currently runs the same tests against all versions of the Elastic Stack.  This requires compatibility code this about to get more challenging to maintain, as the next major release makes significant changes to the field names of APM data.

Proposal: Continue launching supported versions of the Elastic Stack using `compose.py` but split integration tests to support major versions.  

To accomplish this, maintain a branch per supported major.  `compose.py` on `master` can still launch all versions of the stack but the `6.x` branch would maintain tests that are only compatible with that major, `master` would test against `7.x` (then the same transition would happen around `8.x`).  This means the version matrices will be split by major.  Probably a good thing since exclusions starting to dominate now anyway.

graphaelli:
@kuisathaverat Are you available to pick this up?  I can grab it if not, it's rather urgent as tests against 6.x stacks will fail until we address this.

kuisathaverat:
@graphaelli I will pick it up Tomorrow in my morning, I will make a new branch in the apm-integration-testing compatible with APM version 6.x, so the master would be 7.x compatible until we have 8.x, then I will make a job for this 6.x similar to https://apm-ci.elastic.co/job/elastic+apm-integration-testing+master+periodic/, I suppose that this is the urgent part. finally, we will try to build both stack test in the same way (job) and be prepared for the next stack version.

I will cut the branch in the last successful build https://apm-ci.elastic.co/job/elastic+apm-integration-testing+master+multijob-testsuite/71/testReport/

graphaelli:
That's great news, thank you @kuisathaverat!  I actually branched a `6.x` before merging #263 but feel free to branch elsewhere and force push if needed, though I don't that will be necessary.

kuisathaverat:
I am testing the first version https://github.com/elastic/apm-integration-testing/pull/265, it mostly works, I have to review some tests that are falling before started, and I have to generate an HTML file from the JSON result file. On a second iteration, we can tray to split test on different jobs and put some code in the library for reutilization.

https://apm-ci.elastic.co/blue/organizations/jenkins/apm-server%2Fapm-integration-tests-mbp/detail/PR-265/4/pipeline/23
https://apm-ci.elastic.co/job/apm-server/job/apm-integration-tests-mbp/job/PR-265/4/artifact/results.json

![image](https://user-images.githubusercontent.com/5400788/50986970-8d890680-1508-11e9-95c2-613e83c1f36e.png)

[...]

PR #263, opened 2019-01-09
update for v1 removal

graphaelli:


graphaelli:
sure, been meaning to do that since opening #261.  I'll push up a `6.x` now

PR #264, opened 2019-01-09
add --elasticsearch-data-dir option

graphaelli:
To enable upgrade testing, like:

```
./scripts/compose.py start 6.6 --elasticsearch-data-dir "" ...
# don't remove any volumes here
./scripts/compose.py start 7.0.0 --elasticsearch-data-dir "" ...
```

@beniwohli I'm pretty sure this versioned data dir has been here for a long time, do you recall why or if this is a bad idea?  I hesitate to add yet another option to control this but can do it if there's a risk this will cause people problems.

### DIFF
--- a/scripts/compose.py
+++ b/scripts/compose.py
@@ -590,7 +590,7 @@ class Elasticsearch(StackService, Service):
 
         java_opts_env = "ES_JAVA_OPTS=" + " ".join(["-{}{}".format(k, v) for k, v in sorted(es_java_opts.items())])
         self.environment = self.default_environment + [
-            java_opts_env, "path.data=/usr/share/elasticsearch/data/" + self.version]
+            java_opts_env, "path.data=/usr/share/elasticsearch/data/"]
         if not self.oss:
             self.environment.append("xpack.security.enabled=false")
             self.environment.append("xpack.license.self_generated.type=trial")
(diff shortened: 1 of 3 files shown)


graphaelli:
It occurred to me that it was probably so users didn't have to wipe their data volume between starts.  I wonder though if that's even more confusing since sometimes old data will show up and sometimes it won't (when relaunching the same version later).

beniwohli:
The main reason I introduced versioned directories was to avoid breakage when jumping between major versions. e.g. running 6.x, then master, then 6.x again could break because master updates some system indexes that then don't work anymore on 6.x.

TBH, I think we should retain versioned directories. compose.py is brittle enough as it is, and this would introduce another failure point with hard to understand error messages. I agree that adding _another_ flag is not great. Maybe we have to accept the fact that we can't test _every_ possible scenario under the sun with compose.py.

graphaelli:
Fair enough.  Upgrade testing is really important so I'll switch this to add a new flag and leave the default data dirs versioned.

graphaelli:
done, ready again for review

graphaelli:
jenkins, retest this please

graphaelli:
the test failures seem to be unrelated to this change.

graphaelli:
jenkins, retest this please

graphaelli:
spoke with beniwohli, he's 👍

PR #270, opened 2019-01-15
fix: update chrome and pm2 to fix unhealthy rum instance

vigneshshanmugam:
+ chrome is changed from stable - unstable.. https://github.com/GoogleChrome/puppeteer/blob/master/docs/troubleshooting.md#running-puppeteer-in-docker

+ have updated pm2 to latest because of https://github.com/Unitech/pm2/issues/3266#issuecomment-366917676. 

pm2 daemon was not getting killed after doing these two changes,
+ Memory and CPU was not bound, So if the issue continues to appear we need to either increase the timeout of health check for this specific image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,12 +3,12 @@ FROM python:3.7
 RUN curl -sS -o - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - && \
     echo "deb http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google-chrome.list && \
     apt-get -yqq update && \
-    apt-get -yqq install google-chrome-stable && \
+    apt-get -yqq install google-chrome-unstable && \
     rm -rf /var/lib/apt/lists/*
 RUN curl -SLO https://chromedriver.storage.googleapis.com/$(curl -o- https://chromedriver.storage.googleapis.com/LATEST_RELEASE)/chromedriver_linux64.zip && \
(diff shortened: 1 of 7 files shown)


vigneshshanmugam:
Can anyone help here why the CI is failing? Doesn't seem to be related to the change. @graphaelli @kuisathaverat

kuisathaverat:
The APM server is not started correctly, in those cases, you can relaunch the job to see if it is consistent, I think is not related too. 

```
ERROR: for apm-server  Container "34e0b7b79b03" is unhealthy.
```

vigneshshanmugam:
@kuisathaverat Sure Thanks, I have done a dummy commit to trigger build

graphaelli:
jenkins, retest this please

vigneshshanmugam:
@graphaelli right now it goes to the default path and we are using the default timeout for the healthcheck. Probably we can run healthcheck on another port just to keep the instance running without getting killed by race between load gen and healthcheck on same port.

graphaelli:
jenkins, retest this please

graphaelli:
@vigneshshanmugam now the CI is actually able to run the services, the failures indicate this was branched before a few fixes went in.  I've merged master in and pushed to your branch, I hope you don't mind.  We should squash merge when this is ready to go.

Also, please disregard my comment about the healthcheck, I was referring to the rum agent container (used by integration tests) and this only changes the opbeans rum container.

vigneshshanmugam:
@graphaelli Thanks a lot for taking care of the broken build.. Sure no problem 👍. I will squash and merge the PR now.

PR #275, opened 2019-01-24
fix opbeans-rum healthcheck

graphaelli:
working around need for host header


graphaelli:
I haven't seen an issue with that container taking > 1 minute to start up so let's just watch it and follow up if needed.

PR #287, opened 2019-01-31
Add latest server versions

simitt:
This PR adds the missing 6.x server versions to the test matrix, and respects the same exclude versions for the agents as 6.5 had setup.


simitt:
cc @roncohen

PR #291, opened 2019-02-04
explict name for default network is unsupported in old docker-compose

roncohen:
this sets an explicit network name in each service

### DIFF
--- a/scripts/compose.py
+++ b/scripts/compose.py
@@ -1822,14 +1822,6 @@ class LocalSetup(object):
         )
 
         parser.add_argument(
-            '--secret-token',
-            action='store',
-            dest='secret_token',
-            help='Secret token to authenticate against the APM server. Empty by default.',
-            default=''
-        )
-
-        parser.add_argument(
             '--opbeans-apm-js-server-url',
             action='store',
             help='server_url to use for Opbeans frontend service',
@@ -1870,6 +1862,14 @@ class LocalSetup(object):
         parser.add_argument(
             "--opbeans-frontend-bundle-path",
             help='Bundle path in minified files. Defaults to "http://opbeans-node:3000/static/js/" + name of sourcemap',
+        )
+
+        parser.add_argument(
+            '--secret-token',
+            action='store',
+            dest='secret_token',
+            help='Secret token to authenticate against the APM server. Empty by default.',
+            default=''
         )
 
         return parser
@@ -2063,13 +2063,13 @@ class LocalSetup(object):
             auth_header = ''
         print("Uploading {} for {} version {}".format(sourcemap_file, service_name, service_version))
         cmd = (
-            'curl -sS -X POST '
+            'curl -f -X POST '
             '-F service_name="{service_name}" '
             '-F service_version="{service_version}" '
             '-F bundle_filepath="{bundle_path}" '
             '-F sourcemap=@/tmp/sourcemap '
             '{auth_header}'
-            '{server_url}/v1/client-side/sourcemaps'
+            '{server_url}/assets/v1/sourcemaps'
         ).format(
             service_name=service_name,
             service_version=service_version,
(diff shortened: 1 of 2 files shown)


scripts/compose.py
```diff
@@ -234,6 +234,10 @@ def default_logging():
                 "max-size": "2m"
             }
         }
+    
+    @staticmethod
+    def default_networks():
+        return ["net1"]
```
jalvz: maybe a more descriptive name? :)

roncohen: good point 👍

scripts/compose.py
```diff
@@ -1815,6 +1820,14 @@ def init_start_parser(self, parser, services, argv=None):
             default=DEFAULT_APM_SERVER_URL,
         )
 
+        parser.add_argument(
```
graphaelli: `--apm-server-secret-token` is already an option, should use that instead.

graphaelli: note you'll need to merge master in following #292

roncohen: thanks. I've reverted my changes. There's still the separate option for `--secret-token` for sourcemap uploads as it is. I decided not to try and change that in this PR.

roncohen:
* Changed sourcemap upload URL to work with 7.0. 
* Discovered that Curl would silently fail if the wrong token or wrong URL was used. I changed some of the arguments to curl to make it fail loudly now if upload doesn't work. The regular case (correct token/url) is slightly more noisy, but i think it's worths it: 

Now, correct token:
```
 $ ./scripts/compose.py upload-sourcemap --secret-token abc
Uploading /Users/ron/Projects/elastic/apm-integration-testing/docker/opbeans/node/sourcemaps/main.a69bd995.js.map for client version 1.0.0
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 2383k    0     0  100 2383k      0  15.0M --:--:-- --:--:-- --:--:-- 15.1M
```
Incorrect token:
```
$ ./scripts/compose.py upload-sourcemap --secret-token asd
Uploading /Users/ron/Projects/elastic/apm-integration-testing/docker/opbeans/node/sourcemaps/main.a69bd995.js.map for client version 1.0.0
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0 2383k    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 401 Unauthorized
Traceback (most recent call last):
  File "./scripts/compose.py", line 2235, in <module>
    main()
  File "./scripts/compose.py", line 2231, in main
    setup()
  File "./scripts/compose.py", line 1687, in __call__
    self.args.func()
  File "./scripts/compose.py", line 2083, in upload_sourcemaps_handler
    subprocess.check_output(cmd, shell=True).decode('utf8').strip()
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 573, in check_output
    raise CalledProcessError(retcode, cmd, output=output)
subprocess.CalledProcessError: Command 'docker run --rm --network apm-integration-testing -v /Users/ron/Projects/elastic/apm-integration-testing/docker/opbeans/node/sourcemaps/main.a69bd995.js.map:/tmp/sourcemap centos:7 curl -f -X POST -F service_name="client" -F service_version="1.0.0" -F bundle_filepath="http://opbeans-node:3000/static/js/main.a69bd995.js.map" -F sourcemap=@/tmp/sourcemap -H "Authorization: Bearer asd" http://apm-server:8200/assets/v1/sourcemaps' returned non-zero exit status 22
```

roncohen:
even setting `{"name": "apm-integration-testing"}` doesn't work in older docker-compose versions unfortunately, so i've completely removed it

roncohen:
not sure if we want to merge this. The network name will now be set to`<parent-directory>_default`

kuisathaverat:
@roncohen Is this PR still useful?

PR #301, opened 2019-02-04
update agent/stack 6.x version matrix

graphaelli:
based on https://www.elastic.co/guide/en/apm/get-started/6.6/agent-server-compatibility.html, restrict tests against the latest Elastic Stack to compatible agent versions.

related to #300


kuisathaverat:
Should APM Server 6.5+ (6.6, 6.x) work with old versions of the agents? the compatibility matrix point to no.

simitt:
Yes, APM Server up until 7.0 should work with old agents. But that is not true the other way around - agents that work with >= 6.5 will not work with APM Server < 6.5.

kuisathaverat:
> Since apm-server 6.6 is the latest release, the latest released agents and master should also be tested against 6.x. Otherwise we wouldn't have any integration test coverage for what is currently released (or do I miss anything here?).

In that case, we need to add 6.x, and 6.6 to the https://github.com/elastic/apm-integration-testing/blob/master/tests/versions/apm_server.yml file on the master branch. We would make another branch when we break compatibility again

simitt:
> In that case, we need to add 6.x, and 6.6 to the https://github.com/elastic/apm-integration-testing/blob/master/tests/versions/apm_server.yml file on the master branch.

I disagree. In the master branch we query for the fields at the new locations. This does not work for apm-server 6.x, as the resulting documents look different than in master. In my opinion APM Server 6.6, 6.x needs to be part of integration testing 6.x. We have two possibilities: either we create a separate `6.x-v1` branch where we test the agents that only support v1 with all apm server 6.x versions, or we keep only the `6.x` branch and define the exceptions for v2 and v1 agents.

kuisathaverat:
> we create a separate `6.x-v1` branch where we test the agents that only support v1 with all apm server 6.x versions

I prefer this option at some point we will have to add more and more exceptions until we would have to create the branch v1, for sure

graphaelli:
closing in favor of v1/v2 approach

PR #317, opened 2019-02-06
rudimentary xpack security support

graphaelli:
Introduce a new option `--xpack-secure` to enable xpack security in elasticsearch, requiring authentication for all es operations except cluster monitoring.  Kibana prompts for login except for status checks.  Only kibana and apm-server are automatically to communicate with elasticsearch with a password - that's probably a problem for any opbeans apps that use es.

Provided users are: `admin` `apm_server_user` `apm_user_ro` `kibana_system_user` all with password `changeme`

This doesn't work perfectly with 6.x.  No care was taken to make it work with pre-6.6 versions, yet.
* apm-server is not able to load dashboards, it doesn't try to regardless of `--no-apm-server-dashboards` setting when `--xpack-secure` is set.
```
localtesting_6.6.0_apm-server | Exiting: Error importing Kibana dashboards: fail to import the dashboards in Kibana: Error importing directory /usr/share/apm-server/kibana: Failed to import index-pattern: Failed to load directory /usr/share/apm-server/kibana/6/index-pattern:
localtesting_6.6.0_apm-server |   error loading /usr/share/apm-server/kibana/6/index-pattern/apmserver.json: action [indices:data/write/bulk[s]] is unauthorized for user [apm_server_user]. Response: {"objects":[{"id":"apm-*","type":"index-pattern","error":{"message":"action [indices:data/write/bulk[s]] is unauthorized for user [apm_server_user]"}}]}
```

Fun fact, the elasticsearch healthcheck passes even when it shouldn't, eg this passes:
```
curl -s http://localhost:9200/_cluster/health 
{"error":{"root_cause":[{"type":"security_exception","reason":"action [cluster:monitor/health] is unauthorized for user [_anonymous]"}],"type":"security_exception","reason":"action [cluster:monitor/health] is unauthorized for user [_anonymous]"},"status":403}
```

To set passwords for built-in users, something like this works:
`docker-compose exec elasticsearch bin/elasticsearch-setup-passwords auto`

### DIFF
--- a/scripts/compose.py
+++ b/scripts/compose.py
@@ -394,7 +394,8 @@ class ApmServer(StackService, Service):
 
         if self.options.get("enable_kibana", True):
             self.depends_on["kibana"] = {"condition": "service_healthy"}
-            if options.get("apm_server_dashboards", True) and not self.at_least_version("7.0"):
+            if options.get("apm_server_dashboards", True) and not self.at_least_version("7.0") \
+                    and not self.options.get("xpack_secure"):
                 self.apm_server_command_args.append(
                     ("setup.dashboards.enabled", "true")
                 )
--- a/scripts/compose.py
+++ b/scripts/compose.py
@@ -739,8 +739,8 @@ class Kibana(StackService, Service):
             if self.at_least_version("6.3"):
                 self.environment["XPACK_XPACK_MAIN_TELEMETRY_ENABLED"] = "false"
             if options.get("xpack_secure"):
-                self.environment["ELASTICSEARCH_USERNAME"] = "kibana_system_user"
                 self.environment["ELASTICSEARCH_PASSWORD"] = "changeme"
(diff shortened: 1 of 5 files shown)


graphaelli:
@ph @mattapperson possibly interesting for you

scripts/compose.py
```diff
@@ -650,6 +665,13 @@ def __call__(self, parser, namespace, value, option_string=None):
...
 
     def _content(self):
+        volumes = ["esdata:/usr/share/elasticsearch/data"]
+        if self.xpack_secure:
+            volumes.extend([
+                "./docker/elasticsearch/roles.yml:/usr/share/elasticsearch/config/roles.yml",
+                "./docker/elasticsearch/users:/usr/share/elasticsearch/config/users",
+                "./docker/elasticsearch/users_roles:/usr/share/elasticsearch/config/users_roles",
+            ])
         return dict(
```
ph: File realm is the way to go++ :)

PR #319, opened 2019-02-07
temporarily remove python from all agent test

graphaelli:
as a test


simitt:
@kuisathaverat could you have a look why the pipeline tests differ from the old ci here.

kuisathaverat:
@simitt new CI is testing against ES 6.5, old CI uses ES master, I tested the PR with master and we have the same result.

graphaelli:
We've mostly confirmed there is an issue specific to the python agent.  Merging this in to get these green again and we'll revert once we know more.

PR #339, opened 2019-02-20
remove bootstrap command in rum image

vigneshshanmugam:
+ We dont need this anymore after this change https://github.com/elastic/apm-agent-js-base/pull/185

PR #347, opened 2019-02-25
[6.x] remove bootstrap command in rum image (#339)

graphaelli:
Backports the following commits to 6.x:
 - remove bootstrap command in rum image  (#339)


scripts/compose.py
```diff
@@ -1550,6 +1550,7 @@ class LocalSetup(object):
         '6.5': '6.5.4',
         '6.6': '6.6.1',
         '6.7': '6.7.0',
+        'master': '6.6.1',
```
kuisathaverat: I think is not needed, the build is using the parameter ELASTIC_STACK_VERSION

```suggestion
```

PR #350, opened 2019-02-25
exclude python agent from all agent test

graphaelli:
the new-CI version of #319


kuisathaverat:
This change disables to run Python test on merges, it'll still run Python tests included in the all.sh script on PRs and merges, to disable those test we have to modify all.sh script also to not include Python tests.

graphaelli:
@kuisathaverat that was already done in #319

PR #351, opened 2019-02-26
Add heartbeat

roncohen:
Adds heartbeat and some dummy checks

### DIFF
--- a/scripts/compose.py
+++ b/scripts/compose.py
@@ -733,6 +733,30 @@ class Filebeat(BeatMixin, StackService, Service):
         )
 
 
+class Heartbeat(BeatMixin, StackService, Service):
+    DEFAULT_COMMAND = "heartbeat -e --strict.perms=false"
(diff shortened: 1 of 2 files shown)


scripts/compose.py
```diff
@@ -733,6 +733,30 @@ def _content(self):
         )
 
 
+class Heartbeat(BeatMixin, StackService, Service):
+    DEFAULT_COMMAND = "heartbeat -e --strict.perms=false"
+    docker_path = "beats"
+
+    def __init__(self, **options):
+        del options['enable_kibana']
```
graphaelli: Nice workaround, I'll follow up on this as now multiple beats-based (apm-server & heartbeat) services could use this kind of treatment.

roncohen: "nice" is probably an overstatement, but thanks ;)

roncohen:
looks like CI failures are unrelated. Merging. Thanks for the review!

PR #356, opened 2019-02-27
add --with-services option

graphaelli:
for jamming additional/overwriting entries under "services" within docker-compose.yml

### DIFF
--- a/scripts/compose.py
+++ b/scripts/compose.py
@@ -1885,6 +1885,12 @@ class LocalSetup(object):
         )
 
         parser.add_argument(
+            '--with-services',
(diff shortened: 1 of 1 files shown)


graphaelli:
I'm usually adding services to the apm-integration-testing network for one off testing, trying to avoid making `compose.py` any bigger than it already is.

axw:
@graphaelli Could you create a `docker-compose-override.yml` file instead?

graphaelli:
@axw This doesn't stop you from doing that but I preferred to go with the explicit route for this.  Half way in between would be to call docker-compose with multiple `-f` flags but decided that was also a hassle.

PR #369, opened 2019-03-06
wait for the APM Server up and running

kuisathaverat:
Add APM Server as dependency on every APM agent Docker container, so it will be started before any agent always.


kuisathaverat:
> Works for me since these are only for integration tests and we don't have any coverage on agent behavior while apm-server is unavailable, yet.

Yep, I thought on that also, I think it is something we do not have covered by any test right now

PR #370, opened 2019-03-06
fix(rum): export apm server url before build phase

vigneshshanmugam:
+ This should fix the connection refused error that is happening during the rum integration tests.

### DIFF
--- a/docker/rum/run.sh
+++ b/docker/rum/run.sh
@@ -1,3 +1,1 @@
-export APM_SERVER_URL=$ELASTIC_APM_SERVER_URL 
 npx lerna run serve --stream
-
--- a/scripts/compose.py
+++ b/scripts/compose.py
@@ -930,7 +930,7 @@ class AgentRUMJS(Service):
             logging=None,
             environment={
                 "ELASTIC_APM_SERVICE_NAME": "rum",
-                "ELASTIC_APM_SERVER_URL": self.options.get("apm_server_url", DEFAULT_APM_SERVER_URL),
+                "APM_SERVER_URL": self.options.get("apm_server_url", DEFAULT_APM_SERVER_URL),
             },
             ports=[self.publish_port(self.port, self.SERVICE_PORT)],
         )
--- a/scripts/compose.py
+++ b/scripts/compose.py
@@ -895,7 +895,7 @@ class Zookeeper(Service):
 class AgentRUMJS(Service):
     SERVICE_PORT = 8000
     DEFAULT_AGENT_BRANCH = "master"
-    DEFAULT_AGENT_REPO = "elastic/apm-agent-js-base"
+    DEFAULT_AGENT_REPO = "elastic/apm-agent-rum-js"
 
     def __init__(self, **options):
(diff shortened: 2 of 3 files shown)


vigneshshanmugam:
jenkins test

docker/rum/run.sh
```diff
@@ -1,3 +0,0 @@
-export APM_SERVER_URL=$ELASTIC_APM_SERVER_URL
```
jahtalab: I think the `ELASTIC_APM_SERVER_URL`  is passed during the runtime as well. I know this no longer works since we changed the test scripts on the agent but maybe we should add this capability back to the agent test scripts!

vigneshshanmugam: Sorry Hamid, I did not get it. Isn't both different env variables? AFAIK, we did not change anything on the agent test scripts other than normalising the config.

jahtalab: APM_SERVER_URL was used in the `npm run serve` script to serve a `config.js` file to the tests (via a script tag)! That was produce on the fly before each request! This part was removed in one of the recent PRs.

That way during the docker run time we could pass this env variable and change the apm-server url config.

vigneshshanmugam: Got it now. But the whole refactoring of the config was to remove those kind of runtime flags which was super confusing and also confuses the new contributors(including myself). I wanted to the tests to have config served in a single place and bundling step is the perfect place to keep things in control. 

If we decide to do central configuration management and want to do runtime config reflections, Then it makes sense to think about those use cases. For now I don't see the point of needing a runtime solution. It just complicates the testing part.

jahtalab: Generally speaking, I think it's reasonable to let people to change the APM server url at run time, keeping in mind that we build the image once and might reuse multiple times in different scenarios. With the current solution every time we want to change the server url (e.g. changing the port) we need to rebuild, which is time consuming.

I agree it makes our build set up a bit more complicated but I still think the requirement to change the url at run time is valid.

scripts/compose.py
```diff
@@ -930,7 +930,7 @@ def _content(self):
             logging=None,
             environment={
                 "ELASTIC_APM_SERVICE_NAME": "rum",
-                "ELASTIC_APM_SERVER_URL": self.options.get("apm_server_url", DEFAULT_APM_SERVER_URL),
+                "APM_SERVER_URL": self.options.get("apm_server_url", DEFAULT_APM_SERVER_URL),
```
jahtalab: I think `ELASTIC_APM_SERVER_URL` is a standard among other images!

vigneshshanmugam: We don't even use it anywhere..

kuisathaverat:
jenkins run the tests please

scripts/compose.py
kuisathaverat: on the opbeans projects we use ELASTIC_APM_SERVER_URL

https://github.com/elastic/opbeans-node

vigneshshanmugam: Removed this change

PR #372, opened 2019-03-08
[ruby] Inc queue size, avoid deprecated method call.

simitt:
Set internal queue size from 250 to 2000 to avoid flaky results due to dropping events. Install proper agent version. Use new API and remove deprecated call to `span`.


mikker:
Very good!

Do you think we should set the default max queue size to 2000 in the agent too?

simitt:
I wouldn't do that without knowing how much memory that can take up tbh.

PR #387, opened 2019-03-11
 remove --opbeans-node-service-name option

### LINKED ISSUE #212: Fix custom service name for opbeans-node
With the introduction of #211, it's no longer possible to overwrite the service-name for opbeans-node. 

We still have a config option to do this, but that will not have any any effect:

https://github.com/elastic/apm-integration-testing/blob/2b3b42e054a8fa4a517d7e91a942ae3efe2c9471/scripts/compose.py#L1125

We should therefore either remove it, or find a different way to do this.

graphaelli:
closes #212


graphaelli:
build failure is unrelated

PR #398, opened 2019-03-19
only query elasticsearch / save results for failed tests

graphaelli:
also, don't complain when `/app` doesn't exist for saving these results - that's not good for users outside of CI.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -15,9 +15,13 @@ from tests.fixtures.agents import rum
 
 
 @pytest.hookimpl(hookwrapper=True)
-def pytest_runtest_teardown(item, nextitem):
-    """Called after pytest_runtest_call."""
-    outcome = yield
-    rs = es().es.search(index="apm-*")
-    with open(f'/app/tests/results/data-{item.name}.json', 'w') as outFile:
-        json.dump(rs, outFile, sort_keys=True, indent=4, ensure_ascii=False)
+def pytest_runtest_logreport(report):
+    yield
+    if report.when == "call" and report.failed:
+        rs = es().es.search(index="apm-*", size=1000)
+        name = report.nodeid.split(":",2)[-1]
+        try:
+            with open("/app/tests/results/data-{}.json".format(name), 'w') as outFile:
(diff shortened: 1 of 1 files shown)


conftest.py
```diff
@@ -15,9 +15,13 @@
...
-def pytest_runtest_teardown(item, nextitem):
-    """Called after pytest_runtest_call."""
-    outcome = yield
-    rs = es().es.search(index="apm-*")
-    with open(f'/app/tests/results/data-{item.name}.json', 'w') as outFile:
-        json.dump(rs, outFile, sort_keys=True, indent=4, ensure_ascii=False)
+def pytest_runtest_logreport(report):
+    yield
+    if report.when == "call" and report.failed:
+        rs = es().es.search(index="apm-*", size=1000)
```
kuisathaverat: I have put 10000 in a previous PR, 1000 could not grab all the data

kuisathaverat:
I have used it yesterday to check if a test was failing and it works, the following command gives you the number of events of each type

```
cat data-test_conc_req_rails_foobar.json|jq '.hits.hits[]._source.processor.event'|sort|uniq -c
```

conftest.py
graphaelli: Do we really need all of the documents at that point though?  I expect 10000 could cause some problems on these wimpy machines

kuisathaverat: if you want to check the data stored to see what could happen, you'd need all the data, for example, if you want to check why there are more documents than expected, I saw this error many times and without the data, you cannot figure out what it is the error.

graphaelli: I understand why you'd want that but I"m concerned that it's going to cause problems for these tiny instances.  The tests do cheap count queries whereas this is pulling back the bodies for 1000s of documents - at that point I'd expect to run this test locally and run aggregation queries to figure out what is wrong, not process this debugging file.

kuisathaverat: you will run this when you are wrapping up the test, so you would not affect the test, I have checked it locally and the bigger file is about 15MB

PR #401, opened 2019-03-20
Increase java agent buffer

eyalkoren:
Trying to battle the Java agent failures...

PR #402, opened 2019-03-20
[TEST][WIP] Increase java agent buffer

kuisathaverat:
related to https://github.com/elastic/apm-integration-testing/pull/401

### DIFF
--- a/tests/agent/concurrent_requests.py
+++ b/tests/agent/concurrent_requests.py
@@ -105,7 +105,7 @@ class Concurrent:
         self.logger.debug("Starting tornado I/O loop")
         ioloop.IOLoop.instance().start()
 
-    def check_counts(self, it, max_wait=60, backoff=1):
+    def check_counts(self, it, max_wait=60, backoff=5):
         err = "queried for {}, expected {}, got {}"
 
         def assert_count(terms, expected):
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,9 @@
 **/*.pyc
 .cache
 **/__pycache__
-/.idea
+**/.idea
+**/target
+*.iml
 **/tmp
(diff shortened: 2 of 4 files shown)


tests/agent/concurrent_requests.py
```diff
@@ -105,7 +105,7 @@ def load_test(self):
         self.logger.debug("Starting tornado I/O loop")
         ioloop.IOLoop.instance().start()
 
-    def check_counts(self, it, max_wait=60, backoff=1):
+    def check_counts(self, it, max_wait=60, backoff=5):
```
eyalkoren: @kuisathaverat You only increased the backoff here, not the wait time. I also looked at this but talked to Silvia and got to the conclusion it is not likely. Also, there is always exactly one missing when they fail, which doesn't seem to fit.

eyalkoren:
Failed differently- `AssertionError: queried for [{'processor.event': 'transaction'}], expected 1375, got 0` 😫

tests/agent/concurrent_requests.py
kuisathaverat: I know, I want to not bother Elasticsearch, to see if issues are related to Elasticsearch load, seems not.

kuisathaverat:
I am checking the data https://apm-ci.elastic.co/job/apm-integration-test-axis-pipeline/1799/artifact/tests/results/data-test_concurrent_req_java_spring_foobar.json

kuisathaverat:
There are not transactions, How is it possible?

```
cat tests/results/data-test_concurrent_req_java_spring_foobar.json |jq '.hits.hits[]._source.processor.event'|sort|uniq -c

  12 "metric"
 988 "span"
```

PR #406, opened 2019-03-21
Python don't find latest version

kuisathaverat:
Modifications to make the scrips use the right package


docker/opbeans/python/entrypoint.sh
```diff
@@ -7,7 +7,11 @@ if [[ -f /local-install/setup.py ]]; then
     cd ~/local-install && python setup.py install
     cd -
 elif [[ $PYTHON_AGENT_VERSION ]]; then
+  if [[ "$PYTHON_AGENT_VERSION" != "latest" ]]; then
     pip install -q -U elastic-apm==$PYTHON_AGENT_VERSION
+  else
```
graphaelli: this duplicates the default condition, can this just not be set if the latest is required?

kuisathaverat: sure, let's remove the spaghetti code

PR #408, opened 2019-03-22
consult manifest for build candidate docker image urls

graphaelli:
ran into so many problem with docker images with:
```
 ./scripts/compose.py start 7.0.0-rc1 --bc --with-logstash --apm-server-output=logstash --with-metricbeat
```

### DIFF
--- a/scripts/tests/service_tests.py
+++ b/scripts/tests/service_tests.py
@@ -3,7 +3,6 @@ from __future__ import print_function
 import unittest
 import yaml
 
-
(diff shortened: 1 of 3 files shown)


scripts/tests/service_tests.py
```diff
@@ -209,12 +208,6 @@ def test_agent_java_spring(self):
 
 
 class ApmServerServiceTest(ServiceTest):
-    def test_default_buildcandidate(self):
```
graphaelli: replace with localsetup tests since they required mocking, etc

jalvz:
this works for me :+1:

PR #426, opened 2019-05-01
[APM-CI][Integration Tests] Add as explicit value in the IT choice param

v1v:
It looks like the choice param value list doesn't contain the `UI` option and for some reason, since the jenkins version `2.175` which it was bumped recently, the CI pipelines are broken.


v1v:
Automerge to fix the regression with the `UI` and avoid blocking the ITs

PR #431, opened 2019-05-02
ci(jenkins): add node Docker images to the packer cache

kuisathaverat:
* add Node.js Docker images
* add integration test Docker image


v1v:
For testing purposes only! 

jenkins run the tests

PR #441, opened 2019-05-09
[APM-CI] Testing a specific commit of the apm-agent-java

v1v:


v1v:
It failed now somewhere else, so a bit of flakiness somehow here:
- 
image

I'll close this ticket now as the flakiness branch is now the one looking after this particular behaviour

PR #432, opened 2019-05-02
ci(jenkins): add support to run Packetbeat to grab network traffic info during integration test

kuisathaverat:
* new --with-packetbeat argument, it would run a Docket container executing Packetbeat, this container is attached to the APM Server Docker container network drive.
* use [elasticdump](https://www.npmjs.com/package/elasticdump) to grab documents from Elasticsearch on a test failure.

PR #452, opened 2019-05-23
start beats with --all

graphaelli:
and move move `--all` to `--all-opbeans`


graphaelli:
@kuisathaverat we should coordinate this with #432 as the tests introduced here will fail since `packetbeat` doesn't exist yet.

kuisathaverat:
https://github.com/elastic/apm-integration-testing/pull/432 is ready to merge, Do you want me to merge it before #452?

graphaelli:
yeah, go for it.  I can update this one afterwards.

PR #458, opened 2019-06-03
enable apm-server ILM by default in 7.2+

graphaelli:
adds `--apm-server-ilm-disable` option

### DIFF
--- a/scripts/compose.py
+++ b/scripts/compose.py
@@ -417,6 +417,9 @@ class ApmServer(StackService, Service):
             "enable_elasticsearch", True) else {}
         self.build = self.options.get("apm_server_build")
 
+        if self.at_least_version("7.2") and not self.oss and not self.options.get("apm_server_ilm_disable"):
+            self.apm_server_command_args.append(("apm-server.ilm.enabled", "true"))
+
         if self.options.get("enable_kibana", True):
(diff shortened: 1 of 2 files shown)


scripts/compose.py
```diff
@@ -417,6 +417,9 @@ def __init__(self, **options):
             "enable_elasticsearch", True) else {}
         self.build = self.options.get("apm_server_build")
 
+        if self.at_least_version("7.2") and not self.oss and not self.options.get("apm_server_ilm_disable"):
+            self.apm_server_command_args.append(("apm-server.ilm.enabled", "true"))
```
simitt: Could you also set the `setup.template.overwrite: true` option? If not and the templates have been loaded already for this version, it wouldn't have the expected effect.

graphaelli: My only objection to this is that it wouldn't be possible to disable template overwrite, say if we had an integration test that wanted to that.  As is, gets the behavior you're after  `--apm-server-opt=setup.template.overwrite=true`, but is clearly less friendly for casual compose.py users.  For that reason I'd prefer not to set template overwrite on by default until we make it easy to disable it (and I'd rather not add yet another command line option for that).

simitt: ok, let's go with this then and see if we need any future changes or not.

PR #472, opened 2019-06-17
Implement --opbeans-java-agent-branch option

felixbarny:
This lets you start Opbeans with a specific Java agent branch built from source

It adds very little overhead when not specifying a branch

### DIFF
--- a/docker/opbeans/java/Dockerfile
+++ b/docker/opbeans/java/Dockerfile
@@ -13,7 +13,7 @@ RUN ./build-agent.sh "${JAVA_AGENT_REPO}" "${JAVA_AGENT_BRANCH}"
 
 FROM opbeans/opbeans-java:latest
 # replaces the /agent/elastc-apm-agent.jar if it has been built by build-agent.sh
-COPY --from=0 /agent/* /app
+COPY --from=0 /agent/* /app/
 COPY entrypoint.sh /app/entrypoint.sh
 CMD java -javaagent:/app/elastic-apm-agent.jar -Dspring.profiles.active=customdb\
                                         -Dserver.port=${OPBEANS_SERVER_PORT:-3002}\
--- a/docker/opbeans/java/Dockerfile
+++ b/docker/opbeans/java/Dockerfile
@@ -1,5 +1,19 @@
-FROM opbeans/opbeans-java:latest
+FROM maven:3.5.3-jdk-10
+# setting --build-arg JAVA_AGENT_BRANCH=<branch> causes the agent to be built from source
+# instead of using the agent which comes pre-built with opbeans-java:latest
+ARG JAVA_AGENT_REPO=elastic/apm-agent-java
+ARG JAVA_AGENT_BRANCH=
 
+RUN mkdir /agent \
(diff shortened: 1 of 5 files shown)


docker/opbeans/java/Dockerfile
```diff
@@ -1,10 +1,24 @@
...
+RUN mkdir /agent \
+  # making sure there is at least one file to COPY (otherwise docker complains)
+  && touch /agent/ignore
+COPY build-agent.sh .
+# noop if JAVA_AGENT_BRANCH is not set
+RUN ./build-agent.sh "${JAVA_AGENT_REPO}" "${JAVA_AGENT_BRANCH}"
 
+FROM opbeans/opbeans-java:latest
+# replaces the /agent/elastc-apm-agent.jar if it has been built by build-agent.sh
+COPY --from=0 /agent/* /app
```
graphaelli: This caused an error during build for me:

```
Step 8/11 : COPY --from=0 /agent/* /app
ERROR: Service 'opbeans-java' failed to build: When using COPY with more than one source file, the destination must be a directory and end with a /
```

docker/opbeans/java/Dockerfile
```diff
@@ -1,10 +1,24 @@
-FROM opbeans/opbeans-java:latest
+FROM maven:3.5.3-jdk-10
+# setting --build-arg JAVA_AGENT_BRANCH=<branch> causes the agent to be built from source
+# instead of using the agent which comes pre-built with opbeans-java:latest
+ARG JAVA_AGENT_REPO=elastic/apm-agent-java
+ARG JAVA_AGENT_BRANCH=
```
graphaelli: This is currently always set by `compose.py`, I think changing `"master"` to `""` should be good enough.

https://github.com/elastic/apm-integration-testing/blob/b0cef3414173aa3e5ec8100d1e2ec2b0d810f504/scripts/compose.py#L1730

docker/opbeans/java/Dockerfile
felixbarny: seems to depend on the docker version. it works on my machine. Suffixed `/app/` with a slash, that should work in both cases.

docker/opbeans/java/Dockerfile
felixbarny: done

v1v:
jenkins run the tests please

felixbarny:
I have no idea why this fails 🙇

v1v:
@felixbarny  I took the chance to fix the test with https://github.com/elastic/apm-integration-testing/pull/472/commits/e4e2eab5fe8ee9915586f9bd143c7944271734de 

The issue was when running `make test` which caused the UTs to fail as the default value was removed.

felixbarny:
> I'd rather enforce the master branch within the script as done previously for backward compatiblity

But that means the agent will always be built from source 🤔

v1v:
That's indeed a fact! I'd then revert that change and fix the UTs if you don't mind

felixbarny:
Tests are green!
@v1v as soon as you approve we can merge this I think

PR #476, opened 2019-06-18
fix: allow to set --no-kibana with the --all option

kuisathaverat:
it is not possible to execute `scripts/compose.py start master --all --no-kibana` it fails with the error `ERROR: Service 'heartbeat' depends on service 'kibana' which is undefined.` if you add `--no-heartbeat` it does not have an effect and the docker compose file contains heartbeat, checking the compose.py I've found that the enable_kibana option is removed on the init method, I do not understand why and it works if I remove this line, seems some kind leftover from a test.

### DIFF
--- a/scripts/compose.py
+++ b/scripts/compose.py
@@ -996,6 +996,7 @@ class Heartbeat(BeatMixin, StackService, Service):
     docker_path = "beats"
 
     def __init__(self, **options):
+        options['enable_kibana'] = False
         super(Heartbeat, self).__init__(**options)
         config = "heartbeat.yml"
         self.heartbeat_config_path = os.path.join(".", "docker", "heartbeat", config)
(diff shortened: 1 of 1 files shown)


v1v:
Uhm, it seems it was genuinely added there:
- https://github.com/elastic/apm-integration-testing/pull/351#pullrequestreview-208045580

but i'm afraid I don't have enough context though

graphaelli:
I expect this exists because, in previous versions, beats would fail to start if configured to load dashboards but kibana was unavailable.  The best fix here would be to decouple dashboard loading from kibana enabling but I just don't think it's worth the effort - 

```
options['enable_kibana'] = False
```

should do

kuisathaverat:
ok, I understand it now, I will modify the PR

scripts/compose.py
```diff
@@ -996,7 +996,6 @@ class Heartbeat(BeatMixin, StackService, Service):
     docker_path = "beats"
 
     def __init__(self, **options):
-        del options['enable_kibana']
         super(Heartbeat, self).__init__(**options)
```
```suggestion
        options['enable_kibana'] = False
        super(Heartbeat, self).__init__(**options)
```

PR #479, opened 2019-06-18
feat: change BUILD_OPTS precedence

v1v:
This is caused when BUILD_OPTS and any other env variables are affecting the same flag, so, then BUILD_OPTS got less precedence.

For instance:
- The build for https://github.com/elastic/apm-agent-dotnet/pull/294 was triggered with the values `'BUILD_OPTS', value: "--dotnet-agent-version ${env.GIT_BASE_COMMIT}"` and the ITs were launched with the below setup:

```
python scripts/compose.py start 7.0.0 \
--dotnet-agent-version 83c7570b0eaf1e79852915fa3346c254cbfaa5b4  \
--apm-server-build https://github.com/elastic/apm-server.git@master \
--build-parallel \
--dotnet-agent-version=master \
--no-apm-server-dashboards \
--no-apm-server-self-instrument \
--no-kibana \
--with-agent-dotnet \
--force-build
```

As you can see `--dotnet-agent-version 83c7570b0eaf1e79852915fa3346c254cbfaa5b4` and `--dotnet-agent-version=master` are duplicated and therefore the latest one is the one which it's finally used in this case.

This particular change will change the precedence in the other way around.

### DIFF
--- a/.ci/scripts/dotnet.sh
+++ b/.ci/scripts/dotnet.sh
@@ -9,7 +9,7 @@ if [ -n "${APM_AGENT_DOTNET_VERSION}" ]; then
   EXTRA_OPTS=${APM_AGENT_DOTNET_VERSION/'github;'/'--dotnet-agent-version='}
   EXTRA_OPTS=${EXTRA_OPTS/'release;'/'--dotnet-agent-release='}
   EXTRA_OPTS=${EXTRA_OPTS/'commit;'/'--dotnet-agent-version='}
-  BUILD_OPTS="${BUILD_OPTS} ${EXTRA_OPTS}"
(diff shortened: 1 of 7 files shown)


v1v:
This is not required as the ITs which should be triggered for each apm-agent/server should be the ones for the `All` rather than for each individual compatibliity matrix.

kuisathaverat:
in which case you should set those two arguments? `--dotnet-agent-version 83c7570b0eaf1e79852915fa3346c254cbfaa5b4 and --dotnet-agent-version=master `

v1v:
@kuisathaverat ,  I thought there was another use case, but I realized that only the `All` ITs should be triggered. Therefore this PR is actually misleading, so I closed it as it's not required any further actions

PR #485, opened 2019-06-19
fix(jenkins): withEnv causes ClassCastException in the Pipeline

v1v:
## Highlights
- withEnv is not required when running the script.
- withEnv causes some issues when using the `@Field` annotations.
- Tuned the pipeline to use the same strategy as: https://github.com/elastic/apm-integration-testing/blob/master/.ci/Jenkinsfile


v1v:
Automerge after validating it works as expected:
- See [build](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-integration-tests-selector-mbp/detail/fix%2FEnvActionImpl/2/pipeline)

PR #487, opened 2019-06-19
feat: validate docker-compose has been installed as a prerequisite

### LINKED ISSUE #467: report a proper error when docker compose is not available
The following error is shown when you run compose.py without docker-compose installed, we should report a proper error and add docker-compose as a requirement on the README

```
$ ./scripts/compose.py start master --with-metricbeat --with-filebeat
Starting stack services..

Traceback (most recent call last):
  File "./scripts/compose.py", line 2769, in <module>
    main()
  File "./scripts/compose.py", line 2765, in main
    setup()
  File "./scripts/compose.py", line 2174, in __call__
    self.args.func()
  File "./scripts/compose.py", line 2540, in start_handler
    subprocess.call(docker_compose_cmd + pull_params + image_services)
  File "/usr/lib/python2.7/subprocess.py", line 172, in call
    return Popen(*popenargs, **kwargs).wait()
  File "/usr/lib/python2.7/subprocess.py", line 394, in __init__
    errread, errwrite)
  File "/usr/lib/python2.7/subprocess.py", line 1047, in _execute_child
    raise child_exception
OSError: [Errno 13] Permission denied
```

v1v:
Closes elastic/apm-integration-testing#467

## Example

```
./scripts/compose.py start master --with-metricbeat --with-filebeat
Starting stack services..

ERROR: Docker Compose might be missing. See below for further details.

Traceback (most recent call last):
  File "./scripts/compose.py", line 2815, in <module>
    main()
  File "./scripts/compose.py", line 2811, in main
    setup()
  File "./scripts/compose.py", line 2207, in __call__
    self.args.func()
  File "./scripts/compose.py", line 2581, in start_handler
    raise OSError(err)
OSError: [Errno 2] No such file or directory
```


graphaelli:
Why is this needed?  Wouldn't this prevent a user from even running `--help` without `docker-compose` installed?  I'd prefer we catch the subprocess failures and output helpful messages if that's the goal but would avoid this approach.

v1v:
Indeed, I actually misunderstood the scope of the actual issue, so it makes total sense to throw the whole stack trace and a more meaningful message

scripts/compose.py
```diff
@@ -2566,7 +2561,11 @@ def start_handler(self):
                     docker_compose_build.append("--no-cache")
                 if args["build_parallel"]:
                     docker_compose_build.append("--parallel")
-                subprocess.call(docker_compose_build + build_services)
+                try:
```
mdelapenya: LGTM, but could it be possible to reuse code here? We have three instances of same code

v1v:
@graphaelli , do you think this particular PR provides enough value for the specific use case to detect whether the `docker-compose` binary exists in the PATH? No strong opinions, but just wanna either merge it or close it. Whatever we decide.

Thanks

PR #488, opened 2019-06-20
update to latest apm-server build

graphaelli:
master is now 8.0.0 and uses go 1.12.  We probably need a 7.x branch in this repo akin to 6.x and 6.x-v1, or some other approach for the version differences.

PR #489, opened 2019-06-20
feat: validate docker images using bats

v1v:
## Highlights
- Validate whether the docker images are correct by running some BATS.
- This is a kind of similar approach as we have done so far with the `opbeans`. Use https://github.com/elastic/opbeans-dotnet/pull/5 as a reference.

## What tests are executed
- docker build returns 0
- docker run returns 0
- docker inspect returns Running=True

## Test cases
```
make -C docker test-rum
12-alpine: Pulling from library/node
Digest: sha256:3db913c274aad57981cb25ed59c0b8f99b554789d3cf2a0530d63cbb739aed16
Status: Image is up to date for node:12-alpine
Synchronizing submodule url for 'tests/test_helper/bats-assert'
Synchronizing submodule url for 'tests/test_helper/bats-support'
1..5
ok 1 build image
ok 2 clean test containers
ok 3 create test container
ok 4 test container is running
ok 5 clean test containers
/usr/local/bin/tap-xunit -> /usr/local/lib/node_modules/tap-xunit/bin/tap-xunit
/usr/local/bin/txunit -> /usr/local/lib/node_modules/tap-xunit/bin/tap-xunit
+ tap-xunit@2.4.1
added 21 packages from 21 contributors in 1.637s
```

## Tasks
- [x] Validate apm-server
- [x] Validate opbeans
- [x] Validate agents

@graphaelli , as you are working on https://github.com/elastic/apm-integration-testing/pull/488 I wonder if this particular PR might help you to validate it.


v1v:
List of docker images which require some further manipulation:
- test-agents-nodejs-express
- test-agents-python-django
- test-opbeans-node
- test-opbeans-go

v1v:
For the record, the current test failure is not related to this stage but the UTs and it's a bit random

Jenkinsfile
```diff
@@ -41,25 +41,51 @@ pipeline {
...
-    */
-    stage('Unit Tests'){
-      agent { label 'linux && immutable && docker' }
-      steps {
-        withGithubNotify(context: 'Unit Tests', tab: 'tests') {
-          deleteDir()
-          unstash 'source'
-          dir("${BASE_DIR}"){
-            sh "make test-compose lint"
+    stage('Run Tests') {
```
kuisathaverat: you can make it only in the new pipeline

v1v:
@graphaelli , what do you think about this PR? Does it provide enough value to be merged? Or do you want me to clarify what I tried to solve?

Thanks :)

docker/Makefile
```diff
@@ -0,0 +1,31 @@
+.PHONY: help
+.DEFAULT_GOAL := help
+
+LTS_ALPINE ?= 12-alpine
+
+help: ## Display this help text
+	@grep -E '^[a-zA-Z_-]+[%]?:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-20s\033[0m %s\n", $$1, $$2}'
+
+bats: ## Install bats in the project itself
+	@git clone https://github.com/sstephenson/bats.git
```
graphaelli: this is scary from a security perspective, should be pinned to a commit / using our fork

kuisathaverat:
@Gil about your concern of what those test test :) , right now in this repository we have infrastructure provision (docker-compose) and our integration test, we are running the infrastructure provision part test (compose.py lint, test setup, ...) mixed with our real integration test, we have plans to split this and extend apm-integration-testing to whole observability.

docker/Makefile
v1v: I'll ensure we use a specific commit hash

v1v:
@graphaelli 

The main idea of this is to "easily" detect whether the project is healthy enough to be used when running the ITs at runtime. For sure those issues should happen when running the ITs, but it's more efficient to validate the `infrastructure` in advance and fail the build in case any genuine failures which are detected with the BATS framework.

What do I mean about infrastructure?
- All the dockerfiles which are stored in https://github.com/elastic/apm-integration-testing/tree/master/docker

Answering your questions:

> I'm not clear on what do these tests do though? 

These tests validate each Dockerfile in the docker folder. How? Running docker build, docker run and docker inspect. Then will evaluate the errorlevel and other bits and pieces.

> If the agent apps don't build, don't the integration tests fail already? 

Indeed, but it's a waste of time to allocate workers and wait for the whole python script/compose.py to return the error. In the other hand, this particular BATS validation will happen at the very end of the build.

> Do the opbeans container tests already run in each opbeans repo?

Indeed, but this PR does evaluate all the Dockerfiles, the opbeans ones might have some extra configuration on the top therefore it's a good idea to validate whether this `infra` is healthy enough to run the python compose.py script.

There are a few use cases where we could potentially evaluate with this framework:

1) When bumping the version of the apm-server to another version:
- https://github.com/elastic/apm-integration-testing/blob/master/docker/apm-server/Dockerfile#L1

If there is a typo, we won't detect those issues unless we explicitly test this out.

2) When a new configuration change either in the apm agents or opbeans


> might  I think it's ok to run them all here if not but probably best not to duplicate those?

IMO, I'd rather prefer to have them duplicated as they do evaluate different dockerfiles.

Please let me know if I missed something or you got any further questions.

Tx

graphaelli:
Thanks @v1v, that makes sense now and I agree it provides a good amount of value.

v1v:
I'll merge this particular PR as soon as this [pipeline](https://apm-ci.elastic.co/job/apm-integration-tests/view/change-requests/job/PR-489/) gets green

docker/Makefile
v1v: See https://github.com/elastic/apm-integration-testing/pull/489/commits/a1c8a5f925f7babfa71e6c0a3cd277b422eb3534

v1v:
I actually won't merge anything until https://github.com/elastic/apm-pipeline-library/pull/112 is done, the reason. The docker test stage takes about 25 minutes and I think it should not run per PR unless there are changes in the docker folder.

@kuisathaverat , what do you think if we enable that dynamism, therefore, this stage will be disabled unless the changelist contains any references to the docker files and if so it will evaluate that particular stage as suggested inniatilly in this PR?

kuisathaverat:
I will say more we have to split the infrastructure test pipeline test and integration test pipeline in independent checks, since the infrastructure test are not a simply linter anymore.

v1v:
https://github.com/elastic/apm-integration-testing/pull/489/commits/ccf88994a95557b9a0192ef1de5aec7b7f917630 is now the feature to enable the new `docker tests stage` only when the regexp ^docker/ matches with any of the changes in the change list.

.ci/Jenkinsfile
```diff
@@ -64,36 +68,66 @@ pipeline {
...
-            'Python': {runJob('Python')},
-            'Ruby': {runJob('Ruby')},
-            'RUM': {runJob('RUM')},
-            'UI': {runJob('UI')}
-            ]
+        stage('Docker Tests'){
+          agent { label 'linux && immutable && docker' }
+          when {
+            beforeAgent true
+            expression { return env.DOCKER_UPDATED != "false" }
```
v1v:
```suggestion
            expression { return env.DOCKER_UPDATED == "true" }
```

kuisathaverat:
we'd move the `stage('Docker Tests')` to one of our pipelines to build Docker images, in that way we do not add time to the integration test. We will test all the integration test Docker images daily and we add them to the packer cache, this will increase the speed of the integration test builds.

v1v:
As agreed I just removed the changes in the Jenkinsfile.

v1v:
I was somehow forced from the GitHub UI to do a rebase, even though I did several git merges with the `upstream/master`... so that's the reason of the current git commits and force push. :(

PR #490, opened 2019-06-20
fix: signal internal Node.js workload script to disable

watson:
Currently `--no-opbeans-node-loadgen` doesn't have any effect on opbeans-node as it's not using the regular loadgen container.

The idea of the PR is to repurpose this config option and instead use it to set an environment variable which the internal loadgen script can see and know if it should run or not.

### DIFF
--- a/scripts/compose.py
+++ b/scripts/compose.py
@@ -1855,6 +1855,7 @@ class OpbeansNode(OpbeansService):
                 "ELASTIC_APM_SOURCE_LINES_SPAN_LIBRARY_FRAMES",
                 "WORKLOAD_ELASTIC_APM_APP_NAME=workload",
                 "WORKLOAD_ELASTIC_APM_SERVER_URL={}".format(self.apm_server_url),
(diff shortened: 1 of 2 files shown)


watson:
@v1v Do you know why the tests are failing? They seem unrelated, but I'm not sure

v1v:
It seems a test failure in the downstream job:
- https://apm-ci.elastic.co/blue/organizations/jenkins/apm-integration-test-axis-pipeline/detail/apm-integration-test-axis-pipeline/736/tests

Is that what you meant?

watson:
Yes. So it's not related to this PR? I just wasn't sure if we safely merge this or if it would break master

v1v:
jenkins run the tests please

v1v:
I just added some changes in the UTs to fix that issue (see https://github.com/elastic/apm-integration-testing/pull/490/commits/21c92526ed654a82e4037903f17a609d61b732fc). Besides, I raised https://github.com/elastic/apm-integration-testing/pull/493 to help to analise this kind of issues in the future.

PR #491, opened 2019-06-20
Improve output of Opbeans Node.js entrypoint script

watson:
When installing from the local folder, it currently seems like nothing is happening because there isn't much output. This improves the experience for when you tail the logs so you can actually see which commands are being run.


v1v:
jenkins run the tests

watson:
jenkins run the tests

PR #492, opened 2019-06-20
Update path for APM plugin in repo

dgieselaar:
The current x-pack plugins, including APM, got moved into a new folder (see https://github.com/elastic/kibana/pull/39204). This breaks the `validate-ts-interfaces-against-apm-server-sample-docs` script.

This change updates the `clone-kibana` script with the new path.

### DIFF
--- a/scripts/kibana/validate-ts-interfaces-against-apm-server-sample-docs/scripts/clone-kibana.sh
+++ b/scripts/kibana/validate-ts-interfaces-against-apm-server-sample-docs/scripts/clone-kibana.sh
@@ -5,5 +5,5 @@ echo "Cloning Kibana: $OWNER:$BRANCH"
 
 cd ./tmp
 git clone --depth 1 -b $BRANCH https://github.com/$OWNER/kibana.git
-mv ./kibana/x-pack/plugins/apm/typings/es_schemas ./apm-ui-interfaces
+mv ./kibana/x-pack/legacy/plugins/apm/typings/es_schemas ./apm-ui-interfaces
 rm -rf kibana


scripts/kibana/validate-ts-interfaces-against-apm-server-sample-docs/scripts/clone-kibana.sh
```diff
@@ -5,5 +5,5 @@ echo "Cloning Kibana: $OWNER:$BRANCH"
 
 cd ./tmp
 git clone --depth 1 -b $BRANCH https://github.com/$OWNER/kibana.git
-mv ./kibana/x-pack/plugins/apm/typings/es_schemas ./apm-ui-interfaces
+mv ./kibana/x-pack/legacy/plugins/apm/typings/es_schemas ./apm-ui-interfaces
```
kuisathaverat: we have to maintain the current behavior and support the new one

```suggestion
mv ./kibana/x-pack/plugins/apm/typings/es_schemas ./apm-ui-interfaces \
  || mv ./kibana/x-pack/legacy/plugins/apm/typings/es_schemas ./apm-ui-interfaces

```

dgieselaar: @kuisathaverat 	We're all expected to update our branches ASAP with the new folder paths. Usually this job runs only on `master`, which already has the new paths. Do you think it's worth it for the rare occasion where somebody wants to run this script on a branch that has old paths?

kuisathaverat: I've double checked it, we do not run this test with any other versions than the master, so it is fine to replace the command as you suggested.

kuisathaverat:
jenkins run the tests

PR #493, opened 2019-06-21
ci(jenkins): enable unit tests for the integration tests app

v1v:
## Highlights
- Add a new stage in the pipeline to validate the UTs.
- Therefore it will fail fast if the UTs/Lints are broken rather than waiting for the ITs to fail.


v1v:
jenkins run the tests

v1v:
jenkins run the tests

PR #494, opened 2019-06-21
Remove asserts for stacktrace_frame optional properties

SergeyKleyman:
Remove asserts expecting `function`, `abs_path` and `line` properties in `stacktrace_frame` because according to https://github.com/elastic/apm-server/blob/master/docs/spec/stacktrace_frame.json `filename` is the only required `stacktrace_frame` property.


kuisathaverat:
jenkins run the tests

v1v:
jenkins run the tests

PR #495, opened 2019-06-21
fix: set default values for the rum agent arguments

v1v:
This will help to enable the rum docker image validation with https://github.com/elastic/apm-integration-testing/pull/489


v1v:
jenkins run the tests

PR #503, opened 2019-06-25
fix: persist more builds

kuisathaverat:
* persist more builds
* send notification on cleanup stage
* add some Docker images to the packer cache

### DIFF
--- a/.ci/integrationTestSelector.groovy
+++ b/.ci/integrationTestSelector.groovy
@@ -137,7 +137,7 @@ pipeline {
         unstash "source"
         dir("${BASE_DIR}"){
           script {
-            docker.image('node:11').inside() {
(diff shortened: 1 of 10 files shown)


.ci/integrationTestSelector.groovy
```diff
@@ -137,7 +137,7 @@ pipeline {
         unstash "source"
         dir("${BASE_DIR}"){
           script {
-            docker.image('node:11').inside() {
+            docker.image('node:12').inside() {
```
v1v: This is required from now on? I mean, `11` is not supported anymore?

kuisathaverat: talking with Thomas the other day, told me even versions are the stables, and odd versions are develop versions, so let's use a stable one are develop versions

PR #506, opened 2019-06-25
fix: dict is now ordered by key

v1v:
Some UTs are failing from time to time even though the dict was sorted:
- https://github.com/elastic/apm-integration-testing/pull/498

After a few iterations, I managed to find that the python in use was the one from the host rather than using the docker container which it's created when running the make goal `dockerized-test`. Therefore the problem was an environmental issue.

## Highlights
- Enable UTs using a script to encapsulate how to run it within the CI.
- Ensure the dict is ordered

### DIFF
--- a/scripts/compose.py
+++ b/scripts/compose.py
@@ -2083,13 +2083,12 @@ class OpbeansLoadGenerator(Service):
                         self.loadgen_rpms[service_name.replace('_', '-')] = rpm
 
     def _content(self):
-        keys_order_loadgen_rpms = OrderedDict(sorted(self.loadgen_rpms.items()))
         content = dict(
             image="opbeans/opbeans-loadgen:latest",
             depends_on={service: {'condition': 'service_healthy'} for service in self.loadgen_services},
             environment=[
                 "OPBEANS_URLS={}".format(','.join('{0}:http://{0}:3000'.format(s) for s in self.loadgen_services)),
-                "OPBEANS_RPMS={}".format(','.join('{}:{}'.format(k, v) for k, v in keys_order_loadgen_rpms.items()))
+                "OPBEANS_RPMS={}".format(','.join('{}:{}'.format(k, v) for k, v in sorted(self.loadgen_rpms.items())))
             ],
             labels=None,
         )
--- a/scripts/compose.py
+++ b/scripts/compose.py
@@ -2083,12 +2083,13 @@ class OpbeansLoadGenerator(Service):
                         self.loadgen_rpms[service_name.replace('_', '-')] = rpm
 
     def _content(self):
+        keys_order_loadgen_rpms = OrderedDict(sorted(self.loadgen_rpms.items()))
         content = dict(
             image="opbeans/opbeans-loadgen:latest",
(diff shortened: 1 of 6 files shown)


scripts/tests/localsetup_tests.py
```diff
@@ -403,21 +403,21 @@ def test_opbeans_loadgen(self):
...
-                    opbeans-ruby: {condition: service_healthy}
-                environment:
-                 - 'OPBEANS_URLS=opbeans-python:http://opbeans-python:3000,opbeans-ruby:http://opbeans-ruby:3000'
-                 - 'OPBEANS_RPMS=opbeans-python:50,opbeans-ruby:10'
-                logging:
-                    driver: json-file
-                    options: {max-file: '5', max-size: 2m}""")
+        ).render()["opbeans-load-generator"]
+
+        self.assertEqual("opbeans/opbeans-loadgen:latest", opbeans_load_gen["image"])
```
beniwohli: you can just use 

    assert "opbeans/opbeans-loadgen:latest" == opbeans_load_gen["image"]

and similar for the other asserts

v1v: Awesome, thanks, I'll update it

graphaelli:
Another option is to actually sort the lists in:

https://github.com/elastic/apm-integration-testing/blob/e9e848451e854b7c0f328e282da7f622dadf8122/scripts/compose.py#L2090-L2091

> the dict was sorted

It was actually sorted by insertion order, not sorted by name as you might have expected.

> I hate these tests :)

I feel that but they do provide a lot more confidence when making changes to this now important piece of our testing and development infrastructure.

v1v:
> It was actually sorted by insertion order, not sorted by name as you might have expected.
aha!! I actually thought it works out of the box without needing to do anything else. 😢 
I'll then revert my changes in this PR and do the ordering correctly.

Does it sound good to you?

scripts/compose.py
```diff
@@ -2083,12 +2083,13 @@ def __init__(self, **options):
...
 
     def _content(self):
+        keys_order_loadgen_rpms = OrderedDict(sorted(self.loadgen_rpms.items()))
         content = dict(
             image="opbeans/opbeans-loadgen:latest",
             depends_on={service: {'condition': 'service_healthy'} for service in self.loadgen_services},
             environment=[
                 "OPBEANS_URLS={}".format(','.join('{0}:http://{0}:3000'.format(s) for s in self.loadgen_services)),
-                "OPBEANS_RPMS={}".format(','.join('{}:{}'.format(k, v) for k, v in self.loadgen_rpms.items()))
+                "OPBEANS_RPMS={}".format(','.join('{}:{}'.format(k, v) for k, v in keys_order_loadgen_rpms.items()))
```
graphaelli: I would skip creating a dict for this

```suggestion
                "OPBEANS_RPMS={}".format(','.join('{}:{}'.format(k, v) for k, v in sorted(self.loadgen_rpms.items())))
```

graphaelli:
Sounds good, sorry I wasn't following more closely with the earlier change

PR #507, opened 2019-06-25
ci: enable new pipeline on 6.x branch

kuisathaverat:
enable new pipeline on 6.x branch


v1v:
Although it seems the UTs are failing :(

PR #508, opened 2019-06-25
WiP ci(jenkins): enable multiple apps per agent

v1v:
## Highlights
- Python agent got two different test apps, so let's enable them to be launched too.


v1v:
Although this particular PR could be easily fixed within the ITs call as long as the BUILD_OPTS en variable contains `--with-agent-python-flask`.

v1v:
It seems the python app somehow requires to run both agents, so already fixed with:
- https://github.com/elastic/apm-agent-python/pull/501/commits/06225120602af578ec7866969fcc48968d6810d0

Closing this ticket as it's not required though

PR #510, opened 2019-06-26
ci(jenkins): use agentMapping step

v1v:
## Highlights
- Use the `agentMapping ` step which already contains methods to those maps.

## Tasks
- [x] Backport to the other branches. As agreed we won't backport it
- [x] Revert https://github.com/elastic/apm-integration-testing/pull/510/commits/f2a5485b4cadc8152d5fac1ecec0bd5cb0352cf9

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -1,6 +1,6 @@
 #!/usr/bin/env groovy
 
-@Library('apm@current') _
+@Library('apm@test-itsmaps') _
 
 pipeline {
   agent none
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -52,7 +52,7 @@ pipeline {
           deleteDir()
           unstash 'source'
           dir("${BASE_DIR}"){
-            sh "make test-compose lint"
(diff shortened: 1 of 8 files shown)


kuisathaverat:
@v1v we can keep old branches as they are, I think we no need to backport this

mdelapenya:
Is `its` a good name? I would choose a more representative one

v1v:
What name? I chose that one as we normally use `ITs`, so it was a good candidate, but happy to rename it

mdelapenya:
The `its` step is simply selecting the argument/fields for agents, right? like an agent mapping/facade/selector...

image

mdelapenya:
I like the name ❤️! What I say weird is repeating the word `agent` in the method name. Take a look at those methods:

`agentMapping.ymlFiles` Vs `agentMapping.agentYamlVar` or `agentMapping.mapAgentsIDs`

What do you think of renaming the methods to:

```diff
- agentMapping.mapAgentsIDs
+ agentMapping.mapIDs

- agentMapping.agentYamlVar
+ agentMapping.ymlVars
```

v1v:
jenkins run the tests please

PR #277, opened 2019-01-28
7.0 upgrade test

graphaelli:
adds `--apm-server-tee` option to enable a few hardcoded options for testing upgrades to 7.0

run `scripts/ci/7.0-upgrade.sh` to try it out.  run `./venv/bin/pytest tests/server/test_upgrade.py -vv --showlocals` to iterate quickly.

PR #518, opened 2019-06-28
ci(jenkins): enable upgrade tests pipeline

v1v:
## Highlights
- Enable MBP per branch for running the upgrade tests.
- Runs from Monday to Friday every morning. (CEST)
- It relates to https://github.com/elastic/apm-integration-testing/pull/277
- As agreed we only require this particular pipeline in the `7.x` branch, but this pipeline is ready to be used for other branches if required in the future.

## Note
- I cannot test the pipeline in apm-ci until merged as the `JJBB` doesn't allow, afaik, to generate jobs from PRs.


graphaelli:
This will be great to have but I think it only makes sense on the `7.x` branch - we don't have a 7->8 upgrade test built yet.

v1v:
@graphaelli ,  PR is now based on `7.x`, thanks for the heads up :)

PR #523, opened 2019-07-05
Adapt default handling to apm-server 7.3

simitt:
* Only enable ilm in 7.2, as from 7.3 on it should be set to `auto` by the server's default settings.
* Enable `apm-server.kibana.*` settings by default, provide `--apm-server-acm-disable`flag for disabling it.

cc @elastic/apm-server

### DIFF
--- a/scripts/compose.py
+++ b/scripts/compose.py
@@ -379,6 +379,7 @@ class ApmServer(StackService, Service):
     DEFAULT_MONITOR_PORT = "6060"
     DEFAULT_OUTPUT = "elasticsearch"
     OUTPUTS = {"elasticsearch", "kafka", "logstash"}
+    DEFAULT_KIBANA_HOST = "kibana:5601"
 
     def __init__(self, **options):
         super(ApmServer, self).__init__(**options)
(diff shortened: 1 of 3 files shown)


scripts/compose.py
```diff
@@ -611,6 +620,12 @@ def add_arguments(cls, parser):
             default=[],
             help="arbitrary additional configuration to set for apm-server"
         )
+        parser.add_argument(
+            "--apm-server-disable-acm",
```
graphaelli: can you flip this to `apm-server-acm-disable` to match `--apm-server-ilm-disable` ?

scripts/compose.py
```diff
@@ -418,9 +418,18 @@ def __init__(self, **options):
...
+            self.apm_server_command_args.append(("apm-server.ilm.enabled", "false"))
+        elif self.at_least_version("7.2") and not self.at_least_version("7.3") and not self.oss:
             self.apm_server_command_args.append(("apm-server.ilm.enabled", "true"))
 
+        if self.options.get("disable_acm"):
+            self.apm_server_command_args.append(("apm-server.kibana.enabled", "false"))
+        elif self.at_least_version("7.3"):
+            self.apm_server_command_args.extend([
+                ("apm-server.kibana.enabled", "true"),
+                ("apm-server.kibana.host", "kibana:5601")])
```
graphaelli: let's extract `kibana:5601` and use it for `setup.kibana.host` as well - we probably want to add yet another command line option for this, but later

Issue #531, opened 2019-07-10
Service 'agent-rumjs' failed to build


v1v:
```
19:52:29  Service 'agent-rumjs' failed to build: The command '/bin/sh -c npm install' returned a non-zero code: 1
```

```
19:52:26  > fibers@4.0.1 install /rumjs-integration-test/node_modules/fibers
19:52:26  > node build.js || nodejs build.js
19:52:26  
19:52:26  gyp ERR! configure error 
19:52:26  gyp ERR! stack Error: Can't find Python executable "python", you can set the PYTHON env variable.
19:52:26  gyp ERR! stack     at PythonFinder.failNoPython (/usr/local/lib/node_modules/npm/node_modules/node-gyp/lib/configure.js:484:19)
19:52:26  gyp ERR! stack     at PythonFinder.<anonymous> (/usr/local/lib/node_modules/npm/node_modules/node-gyp/lib/configure.js:406:16)
19:52:26  gyp ERR! stack     at F (/usr/local/lib/node_modules/npm/node_modules/which/which.js:68:16)
19:52:26  gyp ERR! stack     at E (/usr/local/lib/node_modules/npm/node_modules/which/which.js:80:29)
19:52:26  gyp ERR! stack     at /usr/local/lib/node_modules/npm/node_modules/which/which.js:89:16
19:52:26  gyp ERR! stack     at /usr/local/lib/node_modules/npm/node_modules/isexe/index.js:42:5
19:52:26  gyp ERR! stack     at /usr/local/lib/node_modules/npm/node_modules/isexe/mode.js:8:5
19:52:26  gyp ERR! stack     at FSReqWrap.oncomplete (fs.js:152:21)
19:52:26  gyp ERR! System Linux 4.15.0-1036-gcp
19:52:26  gyp ERR! command "/usr/local/bin/node" "/usr/local/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "rebuild" "--release"
19:52:26  gyp ERR! cwd /rumjs-integration-test/node_modules/fibers
19:52:26  gyp ERR! node -v v8.16.0
19:52:26  gyp ERR! node-gyp -v v3.8.0
19:52:26  gyp ERR! not ok 
19:52:26  node-gyp exited with code: 1
19:52:26  Please make sure you are using a supported platform and node version. If you
19:52:26  would like to compile fibers on this machine please make sure you have setup your
19:52:26  build environment--
[...]

PR #532, opened 2019-07-10
fix: missing dependencies in the rumjs agent

v1v:
Closes https://github.com/elastic/apm-integration-testing/issues/531


v1v:
The docker image seems to be built although the docker container seems to exit without an error but with the below logs

```
lerna notice cli v3.14.1
lerna info versioning independent
lerna success run No packages found with the lifecycle script 'serve'
```

Any hints?

v1v:
Unrelated failure tests in the go agent

PR #540, opened 2019-07-15
feat: delete legacy integration testing jobs

kuisathaverat:
* delete legacy integration testing jobs


v1v:
Missing branch in the MBP:
- https://apm-ci.elastic.co/job/apm-integration-test-downstream/

kuisathaverat:
@v1v I think you check the wrong job this is for master that exists, the only branches do not exist  are 6.x and 6.x-v1

PR #542, opened 2019-07-15
feat: delete legacy integration testing jobs

kuisathaverat:
* delete legacy integration testing jobs


v1v:
I guess the pipeline got broken as this is a chicken-egg problem

kuisathaverat:
I know, the file need for this job is in this PR chicken-egg problem :) , I can split the PR but I think it doesn't worth, the same will happens on 6.x

PR #544, opened 2019-07-16
tests/agent: search on processor.event

axw:
Some of the searches we have been using will match
breakdown metrics, since they share some common fields
(notably, transaction.name and service.name). Ensure
we search on "processor.event" where we only care about
specific event types.


kuisathaverat:
@axw Should we backport this to 7.x or 6.x?

axw:
@kuisathaverat yes, this should go into 7.x, but doesn't need to go into 6.x.

PR #547, opened 2019-07-18
feat: pre-commit config

kuisathaverat:
This PR adds the pre-commit config, to use you should install pre-commit in the local repository, Follow https://pre-commit.com/#install and execute `pre-commit install` or `pre-commit install  --allow-missing-config` (only if you are working on an old branch) inside the local repo folder. Then you will have a really nice bunch of pre-commit checks 😄 

Screenshot 2019-07-18 at 19 17 13

### DIFF
--- a/.ci/.yamlint.yml
+++ b/.ci/.yamlint.yml
@@ -1,0 +1,10 @@
+extends: default
+
+rules:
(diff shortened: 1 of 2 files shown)


v1v:
yamllint and shellcheck might be required to be installed

PR #557, opened 2019-07-23
feat: Allow run integration tests against any APM Server and Elasticsearch URL

kuisathaverat:
* Allow disabling APM Server, now fails because agents depend on APM Server even do you pass `--no-apm-server`
* Use the ES_URL elasticsearch to dump test data on failed tests
* Allow overwriting Makefile environment vars
* Allow changing elasticsearch URL for OpBeans, Logstash, and Kibana
* Unit tests for the new arguments

After these changes, it is possible to launch the integration test against one of our k8s clusters or an Elastic Cloud instance.

### DIFF
--- a/scripts/compose.py
+++ b/scripts/compose.py
@@ -193,6 +193,9 @@ def add_agent_environment(mappings):
 class Service(object):
     """encapsulate docker-compose service definition"""
 
+    DEFAULT_ELASTICSEARCH_HOSTS = "elasticsearch:9200"
+    DEFAULT_KIBANA_HOST = "kibana:5601"
+
     # is this a side car service for opbeans. If yes, it will automatically
     # start if any opbeans service starts
     opbeans_side_car = False
@@ -381,7 +384,6 @@ class ApmServer(StackService, Service):
     DEFAULT_MONITOR_PORT = "6060"
     DEFAULT_OUTPUT = "elasticsearch"
     OUTPUTS = {"elasticsearch", "kafka", "logstash"}
-    DEFAULT_KIBANA_HOST = "kibana:5601"
 
     def __init__(self, **options):
         super(ApmServer, self).__init__(**options)
@@ -465,9 +467,7 @@ class ApmServer(StackService, Service):
                 q.pop("write")
             self.apm_server_command_args.append(("queue.spool", json.dumps(q)))
 
-        es_urls = self.options.get("apm_server_elasticsearch_urls")
-        if not es_urls:
-            es_urls = ["elasticsearch:9200"]
+        es_urls = []
 
         def add_es_config(args, prefix="output"):
             """add elasticsearch configuration options."""
@@ -480,9 +480,10 @@ class ApmServer(StackService, Service):
                 elif self.options.get("xpack_secure"):
                     args.append((prefix + ".elasticsearch.{}".format(cfg), default_apm_server_creds.get(cfg)))
 
-        add_es_config(self.apm_server_command_args)
+        es_urls = self.options.get("apm_server_elasticsearch_urls") or [self.DEFAULT_ELASTICSEARCH_HOSTS]
 
         if self.apm_server_output == "elasticsearch":
+            add_es_config(self.apm_server_command_args)
             self.apm_server_command_args.extend([
                 ("output.elasticsearch.enabled", "true"),
             ])
@@ -493,10 +494,10 @@ class ApmServer(StackService, Service):
                     ("apm-server.register.ingest.pipeline.enabled", "true"),
                 ])
         else:
+            add_es_config(self.apm_server_command_args, prefix="xpack.monitoring")
             self.apm_server_command_args.extend([
                 ("output.elasticsearch.enabled", "false"),
             ])
-            add_es_config(self.apm_server_command_args, prefix="xpack.monitoring")
             if self.apm_server_output == "kafka":
                 self.apm_server_command_args.extend([
                     ("output.kafka.enabled", "true"),
@@ -557,7 +558,7 @@ class ApmServer(StackService, Service):
             '--apm-server-elasticsearch-url',
             action="append",
             dest="apm_server_elasticsearch_urls",
-            help="apm-server elasticsearch output url(s) (e.g. 'http://es1.example.com,http://es2.example.com')."
+            help="apm-server elasticsearch output url(s)."
         )
         parser.add_argument(
             '--apm-server-elasticsearch-username',
@@ -884,9 +885,8 @@ class BeatMixin(object):
         parser.add_argument(
             "--{}-elasticsearch-url".format(cls.name()),
             action="append",
-            dest="apm_server_elasticsearch_urls",
-            help="{} elasticsearch output url(s) (e.g. 'http://es1.example.com,http://es2.example.com')."
-                 .format(cls.name()),
+            dest="{}_elasticsearch_urls".format(cls.name()),
+            help="{} elasticsearch output url(s).".format(cls.name())
         )
         parser.add_argument(
             "--{}-elasticsearch-username".format(cls.name()),
@@ -904,6 +904,7 @@ class BeatMixin(object):
         )
 
     def __init__(self, **options):
+        super(BeatMixin, self).__init__(**options)
         self.command = list(self.DEFAULT_COMMAND)
         self.depends_on = {"elasticsearch": {"condition": "service_healthy"}} if options.get(
             "enable_elasticsearch", True) else {}
@@ -912,12 +913,9 @@ class BeatMixin(object):
             self.depends_on["kibana"] = {"condition": "service_healthy"}
         self.environment = {}
 
-        es_urls = options.get("{}_elasticsearch_urls".format(self.name()))
-        if not es_urls:
-            es_urls = ["elasticsearch:9200"]
-
         def add_es_config(args, prefix="output"):
             """add elasticsearch configuration options."""
+            es_urls = options.get("{}_elasticsearch_urls".format(self.name())) or [self.DEFAULT_ELASTICSEARCH_HOSTS]
             default_beat_creds = {"username": "{}_user".format(self.name()), "password": "changeme"}
             args.append((prefix + ".elasticsearch.hosts", json.dumps(es_urls)))
             for cfg in ("username", "password"):
@@ -950,8 +948,6 @@ class BeatMixin(object):
 
         for param, value in command_args:
             self.command.extend(["-E", param + "=" + value])
-
-        super(BeatMixin, self).__init__(**options)
 
     def build_candidate_manifest(self):
         version = self.version
@@ -1054,8 +1050,8 @@ class Kibana(StackService, Service):
                 self.environment["ELASTICSEARCH_PASSWORD"] = "changeme"
                 self.environment["ELASTICSEARCH_USERNAME"] = "kibana_system_user"
                 self.environment["STATUS_ALLOWANONYMOUS"] = "true"
-        self.es_urls = self.options.get("kibana_elasticsearch_urls") or ["http://elasticsearch:9200"]
-        self.environment["ELASTICSEARCH_URL"] = self.es_urls
+        self.environment["ELASTICSEARCH_URL"] = ",".join(self.options.get(
+            "kibana_elasticsearch_urls") or [self.DEFAULT_ELASTICSEARCH_HOSTS])
 
     @classmethod
     def add_arguments(cls, parser):
@@ -1063,7 +1059,7 @@ class Kibana(StackService, Service):
             "--kibana-elasticsearch-url",
             action="append",
             dest="kibana_elasticsearch_urls",
-            help="kibana elasticsearch output url(s) (e.g. 'http://es1.example.com,http://es2.example.com')."
+            help="kibana elasticsearch output url(s)."
         )
 
     def _content(self):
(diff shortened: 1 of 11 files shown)


scripts/compose.py
```diff
@@ -1673,6 +1712,12 @@ def add_arguments(cls, parser):
             dest=cls.option_name() + '_agent_local_repo',
             help=cls.name() + " local repo path for agent"
         )
+        parser.add_argument(
```
kuisathaverat: I could not use `--opbeans-elasticsearch-url` as a parameter because it is defined on every child class and causes a conflict in names, however, all parameter write the same option so you only have to set it once.

scripts/compose.py
```diff
@@ -1051,6 +1053,19 @@ def __init__(self, **options):
                 self.environment["ELASTICSEARCH_PASSWORD"] = "changeme"
                 self.environment["ELASTICSEARCH_USERNAME"] = "kibana_system_user"
                 self.environment["STATUS_ALLOWANONYMOUS"] = "true"
+        self.es_urls = self.options.get("kibana_elasticsearch_urls")
```
graphaelli: maybe more pythonic to write it as and remove the 2 following lines
```suggestion
        self.es_urls = self.options.get("kibana_elasticsearch_urls") or ["http://elasticsearch:9200"]
```

scripts/compose.py
```diff
@@ -1081,15 +1096,29 @@ def build_candidate_manifest(self):
         return self.bc["projects"]["logstash-docker"]["packages"][key]
 
     def _content(self):
+        self.es_urls = self.options.get("logstash_elasticsearch_urls")
```
graphaelli: same as above:
```suggestion
        self.es_urls = self.options.get("logstash_elasticsearch_urls") or ["http://elasticsearch:9200"]
```

scripts/compose.py
```diff
@@ -1081,15 +1096,29 @@ def build_candidate_manifest(self):
...
+        if not self.es_urls:
+            self.es_urls = ["http://elasticsearch:9200"]
         return dict(
             depends_on={"elasticsearch": {"condition": "service_healthy"}} if self.options.get(
                 "enable_elasticsearch", True) else {},
-            environment={"ELASTICSEARCH_URL": "http://elasticsearch:9200"},
+            environment={
+                "ELASTICSEARCH_URL={}".format(self.es_urls),
```
graphaelli: can this handle multiple urls?  if so, probably need a `",".join(self.es_urls)`, otherwise, maybe just take the first entry?

scripts/compose.py
```diff
@@ -1211,9 +1240,10 @@ def __init__(self, **options):
         super(AgentRUMJS, self).__init__(**options)
         self.agent_branch = options.get("rum_agent_branch", self.DEFAULT_AGENT_BRANCH)
         self.agent_repo = options.get("rum_agent_repo", self.DEFAULT_AGENT_REPO)
-        self.depends_on = {
-            "apm-server": {"condition": "service_healthy"},
-        }
+        if options.get("enable_apm_server", True):
```
graphaelli: is metricbeat ok with the config as-is without an apm-server?  It's possible metricbeat.yml needs some change if no apm-server is going to be available, not certain.

scripts/compose.py
graphaelli: This will probably be a pain.  A workaround is to remember if the option was set and skip it after the first time, like done here: https://github.com/elastic/apm-integration-testing/blob/0d4b122b2b1aaede49a6282083c7e4efb745e346/scripts/compose.py#L1359 (can't just `return` in this case but can guard with an `if`)

scripts/compose.py
```diff
@@ -1728,7 +1773,7 @@ def _content(self):
                 "ELASTIC_APM_FLUSH_INTERVAL=5",
                 "ELASTIC_APM_TRANSACTION_MAX_SPANS=50",
                 "ELASTIC_APM_SAMPLE_RATE=1",
-                "ELASTICSEARCH_URL=http://elasticsearch:9200",
+                "ELASTICSEARCH_URL={}".format(self.es_urls),
```
graphaelli: like the logstash case, needs to be joined or an entry picked

scripts/compose.py
```diff
@@ -1776,7 +1821,7 @@ def _content(self):
...
```
graphaelli: like the logstash case, needs to be joined or an entry picked

scripts/compose.py
```diff
@@ -1841,7 +1886,7 @@ def _content(self):
                 "DATABASE_DIALECT=POSTGRESQL",
                 "DATABASE_DRIVER=org.postgresql.Driver",
                 "REDIS_URL=redis://redis:6379",
-                "ELASTICSEARCH_URL=http://elasticsearch:9200",
+                "ELASTICSEARCH_URL={}".format(self.es_urls),
```
graphaelli: like the logstash case, needs to be joined or an entry picked

scripts/compose.py
```diff
@@ -1964,7 +2009,7 @@ def _content(self):
                 "ELASTIC_APM_SOURCE_LINES_ERROR_LIBRARY_FRAMES",
                 "ELASTIC_APM_SOURCE_LINES_SPAN_LIBRARY_FRAMES",
                 "REDIS_URL=redis://redis:6379",
-                "ELASTICSEARCH_URL=http://elasticsearch:9200",
+                "ELASTICSEARCH_URL={}".format(self.es_urls),
```
graphaelli: like the logstash case, needs to be joined or an entry picked

scripts/compose.py
```diff
@@ -2012,7 +2057,7 @@ def _content(self):
                 "ELASTIC_APM_SERVICE_NAME={}".format(self.service_name),
                 "DATABASE_URL=postgres://postgres:verysecure@postgres/opbeans-ruby",
                 "REDIS_URL=redis://redis:6379",
-                "ELASTICSEARCH_URL=http://elasticsearch:9200",
+                "ELASTICSEARCH_URL={}".format(self.es_urls),
```
graphaelli: like the logstash case, needs to be joined or an entry picked

scripts/compose.py
kuisathaverat: self.es_urls is a String, so in case it has multiple URL it'd be something like this 'url1,url2,url3,url4' so the `.format(self.es_urls)` would put the whole value in the formated string.

scripts/compose.py
kuisathaverat: https://github.com/elastic/apm-integration-testing/pull/557#discussion_r306676939

scripts/compose.py
kuisathaverat: https://github.com/elastic/apm-integration-testing/pull/557#discussion_r306676939

scripts/compose.py
kuisathaverat: https://github.com/elastic/apm-integration-testing/pull/557#discussion_r306676939

scripts/compose.py
kuisathaverat: https://github.com/elastic/apm-integration-testing/pull/557#discussion_r306676939

scripts/compose.py
kuisathaverat: https://github.com/elastic/apm-integration-testing/pull/557#discussion_r306676939

scripts/compose.py
kuisathaverat: if you set `--no-apm-server` means you should have an `apm-server` running in someplace, so you have to have configured `--apm-server-url`

scripts/compose.py
```diff
@@ -1051,6 +1053,19 @@ def __init__(self, **options):
                 self.environment["ELASTICSEARCH_PASSWORD"] = "changeme"
                 self.environment["ELASTICSEARCH_USERNAME"] = "kibana_system_user"
                 self.environment["STATUS_ALLOWANONYMOUS"] = "true"
+        self.es_urls = self.options.get("kibana_elasticsearch_urls")
+        if not self.es_urls:
```
```suggestion
```

scripts/compose.py
```diff
@@ -1051,6 +1053,19 @@ def __init__(self, **options):
...
+            self.es_urls = ["http://elasticsearch:9200"]
```
```suggestion
```

scripts/compose.py
```diff
@@ -1081,15 +1096,29 @@ def build_candidate_manifest(self):
         return self.bc["projects"]["logstash-docker"]["packages"][key]
 
     def _content(self):
+        self.es_urls = self.options.get("logstash_elasticsearch_urls")
+        if not self.es_urls:
```
```suggestion
```

scripts/compose.py
```diff
@@ -1081,15 +1096,29 @@ def build_candidate_manifest(self):
...
+            self.es_urls = ["http://elasticsearch:9200"]
```
```suggestion
```

scripts/compose.py
graphaelli: I don't generally agree with that but I'm ok with imposing that limitation for metricbeat as this is apm-it after all

graphaelli: I still think this may be a problem though: https://github.com/elastic/apm-integration-testing/blob/6ee572b5efbb593204f97427139421727684bb6e/docker/metricbeat/metricbeat.yml#L29

scripts/compose.py
graphaelli: It looks like a list, a test would help demonstrate.

scripts/compose.py
kuisathaverat: finally, I make it global on the `start` parser definition makes more sense.

scripts/compose.py
kuisathaverat: it has a similar issue with the `--no-kibana` https://github.com/elastic/apm-integration-testing/blob/6ee572b5efbb593204f97427139421727684bb6e/docker/metricbeat/metricbeat.yml#L7

kuisathaverat: and here https://github.com/elastic/apm-integration-testing/blob/6ee572b5efbb593204f97427139421727684bb6e/docker/apm-server/Dockerfile#L21

scripts/compose.py
kuisathaverat: I get you now, those options are arrays, not strings

scripts/tests/service_tests.py
```diff
@@ -808,15 +808,15 @@ def test_logstash(self):
 
 class MetricbeatServiceTest(ServiceTest):
     def test_metricbeat(self):
-        metricbeat = Metricbeat(version="6.2.4", release=True).render()
+        metricbeat = Metricbeat(version="6.2.4", release=True, apm_server_pprof_url='apm-server:6060').render()
```
kuisathaverat: render() does not manage argparser default values.

scripts/compose.py
```diff
@@ -1051,6 +1050,17 @@ def __init__(self, **options):
...
                 self.environment["STATUS_ALLOWANONYMOUS"] = "true"
+        self.environment["ELASTICSEARCH_URL"] = ",".join(self.options.get(
+            "kibana_elasticsearch_urls") or [self.DEFAULT_ELASTICSEARCH_HOSTS])
+
+    @classmethod
+    def add_arguments(cls, parser):
+        parser.add_argument(
+            "--kibana-elasticsearch-url",
+            action="append",
+            dest="kibana_elasticsearch_urls",
```
v1v: What's the reason for using an argument as singular-based and a variable as plural-based?

kuisathaverat: kibana_elasticsearch_urls is an array the URLs

kuisathaverat:
yep, I have to add the test because of that it is WIP

scripts/compose.py
```diff
@@ -1051,6 +1050,17 @@ def __init__(self, **options):
                 self.environment["ELASTICSEARCH_PASSWORD"] = "changeme"
                 self.environment["ELASTICSEARCH_USERNAME"] = "kibana_system_user"
                 self.environment["STATUS_ALLOWANONYMOUS"] = "true"
+        self.environment["ELASTICSEARCH_URL"] = ",".join(self.options.get(
+            "kibana_elasticsearch_urls") or [self.DEFAULT_ELASTICSEARCH_HOSTS])
+
+    @classmethod
+    def add_arguments(cls, parser):
+        parser.add_argument(
+            "--kibana-elasticsearch-url",
```
v1v:
```suggestion
            "--kibana-elasticsearch-urls",
```

as stated in https://github.com/elastic/apm-integration-testing/pull/557#discussion_r321641939 ?

scripts/compose.py
```diff
@@ -1081,21 +1091,46 @@ def build_candidate_manifest(self):
...
             healthcheck=curl_healthcheck(9600, "logstash", path="/"),
             ports=[self.publish_port(self.port, self.SERVICE_PORT), "9600"],
             volumes=["./docker/logstash/pipeline/:/usr/share/logstash/pipeline/"]
         )
 
+    @classmethod
+    def add_arguments(cls, parser):
+        super(Logstash, cls).add_arguments(parser)
+        parser.add_argument(
+            "--logstash-elasticsearch-url",
```
v1v:
```suggestion
            "--logstash-elasticsearch-urls",
```

scripts/compose.py
```diff
@@ -2542,6 +2585,13 @@ def init_start_parser(self, parser, services, argv=None):
             default='http://apm-server:8200',
         )
 
+        parser.add_argument(
+            "--opbeans-elasticsearch-url",
```
v1v:
```suggestion
            "--opbeans-elasticsearch-urls",
```

scripts/compose.py
kuisathaverat: it will be incorrect because `--kibana-elasticsearch-url` will set only one URL, but can appear multiple times, it how it was implemented all over the place.

`compose.py 7.2 start --kibana-elasticsearch-url RUL1 --kibana-elasticsearch-url URL2`

v1v: aha!! That's the missing bit I didn't know. Thanks for the clarification :)

scripts/compose.py
kuisathaverat: I've added the proper environment variable with default value on every place it is needed. I've created a unit test for each iteration on {}_elasticsearch_urls and enable_apm_server parameters

kuisathaverat:
ready for review.

PR #563, opened 2019-07-26
feat: update scripts

kuisathaverat:
same as https://github.com/elastic/apm-integration-testing/pull/561 we have to wait for the packer cache to be updated.

### DIFF
--- a/.ci/scripts/shellcheck
+++ b/.ci/scripts/shellcheck
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 set -e
-docker run --rm -v "$PWD:/mnt" koalaman/shellcheck:stable "$@"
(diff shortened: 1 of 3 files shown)


.ci/scripts/shellcheck
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 set -e
-docker run --rm -v "$PWD:/mnt" koalaman/shellcheck:stable "$@"
+docker run --rm -t -v "$PWD:/mnt" -u "$(id -u):$(id -g)" docker.elastic.co/observability-ci/shellcheck "$@"
```
v1v: Not sure whether we could provide this particular preCommit hook within the apm-pipeline-library somehow. WDYT?

kuisathaverat: I thought it was in the library

PR #565, opened 2019-07-26
Add apm-server-proxy (WIP)

sqren:
This adds a service that replaces apm server for testing purposes. Any data sent to apm server will be written to file. This file can later be used to replay event duration test scenarios.

For this to work the generated docker-compose needs to have the following instead of the normal apm server:
```json
    "apm-server-proxy": {
      "build": {
        "context": "docker/apm-server-proxy",
        "dockerfile": "Dockerfile"
      },
      "volumes": [
        "./docker/apm-server-proxy/shared-volume:/usr/app/shared-volume"
      ],
      "command": "npm start",
      "ports": ["127.0.0.1:8200:8200"],
      "environment": [
        "ELASTIC_APM_SERVER_HOST=apm-server-proxy",
        "ELASTIC_APM_SERVER_PORT=8200"
      ],
      "healthcheck": {
        "interval": "10s",
        "retries": 36,
        "test": [
          "CMD",
          "curl",
          "--write-out",
          "'HTTP %{http_code}'",
          "--fail",
          "--silent",
          "--output",
          "/dev/null",
          "http://apm-server-proxy:8200/"
        ]
      }
    },
```

PR #567, opened 2019-07-29
add --apm-server-record option

graphaelli:
for logging payloads sent to apm-server

~for use with~ replaces #565

Sample usage:
```
# launch env
$ ./scripts/compose.py start master --with-opbeans-go --apm-server-record
# export events
$ docker-compose exec apm-server-2 cp /app/events.json .
# watch events live
$ docker-compose exec apm-server-2 tail -f /app/events.json
{"method":"GET","time":"2019-07-30T17:20:40.046920908Z","url":"/config/v1/agents"}
{"body":"{\"metadata\":{\"system\":{\"architecture\":\"amd64\",\"container\":{\"id\":\"5c26fa8518...

```


graphaelli:
@sqren I took some shortcuts to get this working but nothing too offensive to prevent merging.  Please give it a spin when time permits and feel free to merge / push changes as you see fit.

sqren:
fwiw I had to do the following to copy events.json from the container to the host:
```
docker cp localtesting_8.0.0_apm-server-2:/app/events.json .
```

It seems the other command just copies within the container

sqren:
This works well when I only start a single opbeans agent but if I run `--all` it pretty much grinds to a halt and any docker commands take forever (literally: I've not yet see any of them complete yet :D)

graphaelli:
Sorry about the delay and bad instructions @sqren.  You're right, that `docker-compose cp` is busted and your way is right.  I've just pushed a fix for the recorder that was causing the issues you experienced.

sqren:
@graphaelli When I run `--all --apm-server-record` it is because I want to capture data from all agents. I do not care about elasticsearch, apm-server, kibana or any beats.
Would it make sense that by default running `--apm-server-record` would start all agents and `teeproxy` but nothing else? i'd assume that would make the data collection step a bit faster.

graphaelli:
> Would it make sense that by default running --apm-server-record would start all agents and teeproxy but nothing else?

I added a commit to this to allow use of the apm-server file output, so this should do it:

```
./scripts/compose.py start master --apm-server-record --apm-server-output=file --no-kibana --no-elasticsearch --all
```

sqren:
Awesome, thanks!

PR #571, opened 2019-08-02
feat: use the pre-commit hooks library of apm-pipeline-library

v1v:
## Highlights
- Let's consume the pre-commit hooks library.


.pre-commit-config.yaml
```diff
@@ -36,52 +36,15 @@ repos:
         types: [yaml]
         exclude: (^.pre-commit-config.yaml$|^.ci/.yamlint.yml$|^target/)
 
--   repo: local
+-   repo: git@github.com:v1v/apm-pipeline-library
```
v1v: This should be changed once the feature has been merged in this repo

.ci/Jenkinsfile
```diff
@@ -78,8 +78,10 @@ pipeline {
             withGithubNotify(context: 'Sanity checks', tab: 'tests') {
               deleteDir()
               unstash 'source'
-              dir("${BASE_DIR}"){
-                preCommit(commit: "${GIT_BASE_COMMIT}", junit: true)
+              dir(BASE_DIR){
+                sshagent(['f6c7695a-671e-4f4f-a331-acdce44ff9ba']) {
```
mdelapenya: why running the preCommit in an SSH agent?

v1v: sshagent is required as long as https://github.com/elastic/apm-integration-testing/pull/571/files#diff-82c3b36709926de3a1b27bea0d8258ccR39 is a private repo.

Not sure yet whether those implementation details should be defined within the `preCommit` step to keep the pipeline simple or just wait until the private-repo has been set.  That's the reason I have not set ready this PR yet, but wanted to PoC this approach worked in other repos.

.pre-commit-config.yaml
v1v:
```suggestion
-   repo: git@github.com:elastic/apm-pipeline-library
```

.pre-commit-config.yaml
```diff
@@ -36,52 +36,15 @@ repos:
...
+    rev: feature/precommit-lib
```
v1v:
```suggestion
    rev: current
```

.ci/Jenkinsfile
v1v: Already removed from the pipeline itself, as it's now provided within the `preCommit` step then once we open the library we can deprecate that particular behavior without changing any pipelines.

PR #593, opened 2019-09-17
feat: add option to set service.enviroment value for opbeans

kuisathaverat:
it adds new options to set the value of `service.enviroment` for each Opbean. It will also allow launching a second instance of every Opbeans type with his own configuration, this second instance would be only launched is you request it with the `--with-opbeans-<AGENT>01` parameter. these are the new configuration parameter we will have on compose.py

* `--opbeans-<AGENT>-service-environment=<ENVIRONMENT_NAME>` ser the environment name
* `--with-opbeans-<AGENT>01` enable a second agent for <AGENT> type
* `--with-opbeans-<AGENT>01-*` configuration parameters for the second instance

some examples:

```shell
compose.py start 7.3.0 --all
``` 
it would launch every service but opbeans-<AGENTS>01

```shell
compose.py start 7.3.0 --with-opbeans-java --with-opbeans-java01
``` 
it would launch 2 Opbeans Java instances with the default configuration, the service environment will have different values on each opbeans instance, opbeans-java will use `production` and opbeans-java01 will use `testing`

```shell
compose.py start 7.3.0 \
  --with-opbeans-java  --opbeans-java-service-environment=stage\
  --with-opbeans-java01 -opbeans-java01-service-environment=pro
``` 
it would launch two Opbeans Java instances with two service environment values, opbeans-java will use `stage` and opbeans-java01 will use `pro`

### DIFF
--- a/scripts/compose.py
+++ b/scripts/compose.py
@@ -1694,6 +1694,7 @@ class OpbeansService(Service):
         self.opbeans_branch = options.get(self.option_name() + "_branch") or ""
         self.opbeans_repo = options.get(self.option_name() + "_repo") or ""
         self.es_urls = ",".join(self.options.get("opbeans_elasticsearch_urls") or [self.DEFAULT_ELASTICSEARCH_HOSTS])
+        self.service_environment = options.get(self.option_name() + "_service_environment") or ""
 
     @classmethod
     def add_arguments(cls, parser):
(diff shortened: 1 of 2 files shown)


kuisathaverat:
I've to fix a bunch of tests adding the new environment variable 😢

mdelapenya:
Do we just need to run two services of an agent type? If so, the solution seems look to me.

On the other hand I see maintainability issues with this `--with-opbeans-<AGENT>01` pattern, as it allows just two versions of the agent.

I would see convenient a way of running the CLI just adding a new agent with X configuration, which joins the system.

Wdyt?

>Side note: my above comment is not a blocker, it's just a thought for future refactoring...

kuisathaverat:
>Do we just need to run two services of an agent type? If so, the solution seems look to me.

two should be enough for now, let's keep it simple.

>On the other hand I see maintainability issues with this --with-opbeans-<AGENT>01 pattern, as it allows just two versions of the agent.
I would see convenient a way of running the CLI just adding a new agent with X configuration, which joins the system.
Wdyt?

The new classes are pretty simple extends from the base class and overwrite a couple of variables, that's it, I'd try to keep it as simple as possible if in the future we need more instances of the same Opbean we can start thinking on something that provision an Opbean in a dynamic way but this it is too complex right now for what we need.

kuisathaverat:
the test failures are not related to the PR

PR #597, opened 2019-09-19
refactor: split compose.py into modules

kuisathaverat:
## What does this PR do?
This PR refactor the compose.py file into several files, those files contain closed related classes and methods/functions. There are no changes on code after the refactor, only imports and tests to make it work with modules.

## Why is it important?
With only one Python file every change we new feature, bug fix or whatever we modified is tracked on the same file, this means that there is no easy way to see for example what changes we made on the Opbeans containers on the last 2 months, and similar stuff. 
Modify always the same file makes us face conflicts to merge changes every time we marge a PR, this makes difficult to keep PRs updated and many times is a hell to resolve conflicts or we make mistakes when we resolve them.
When you modify code on a 3150 code lines file, you spend most of the time searching for `strings` to locate a piece of code that you want to modify, small files are easy to manage and because they contain only related code it is easy to locate stuff and fix it.

## Notes
Because we have several modules the path where the `compose.py` can be launch it is now forced to the root of the repo so `./scripts/compose.py` is the command that we have to use now, it is not a big deal because it is the one we've documented on the README.

## Author's checklist
- [x] Unit test pass
- [x] Test locally `--all` and `--all-opbeans`


scripts/helpers.py
```diff
@@ -0,0 +1,152 @@
...
+    return {
+        "interval": interval,
+        "retries": retries,
+        "test": ["CMD", "curl", "--write-out", "'HTTP %{http_code}'", "--fail", "--silent",
+                 "--output", "/dev/null",
+                 "http://{}:{}{}".format(host, port, path)]
+    }
+
+
+build_manifests = {}  # version -> manifest cache
```
v1v: Uhm, should it be moved to the method itself?

scripts/tests/localsetup_tests.py
```diff
@@ -8,14 +8,14 @@
...
+from scripts.elastic_stack import ApmServer, Elasticsearch, Kibana
 
-from ..compose import (ApmServer, Kibana, Elasticsearch)
+from scripts.aux_services import Postgres, Redis
 
-from ..compose import (Postgres, Redis)
-
-from ..compose import LocalSetup, discover_services, parse_version, OpbeansService
+from ..compose import LocalSetup
+from scripts.opbeans import OpbeansService, OpbeansDotnet, OpbeansGo, OpbeansJava, OpbeansNode, OpbeansPython, \
```
v1v: typo?

```suggestion
from scripts.opbeans import OpbeansService, OpbeansDotnet, OpbeansGo, OpbeansJava, OpbeansNode, OpbeansPython, 
```

scripts/tests/service_tests.py
```diff
@@ -5,15 +5,12 @@
...
 
-from ..compose import (AgentDotnet, AgentGoNetHttp, AgentJavaSpring, AgentNodejsExpress,
-                       AgentPythonDjango, AgentPythonFlask, AgentRubyRails)
-
-from ..compose import (ApmServer, Kibana, Elasticsearch, Filebeat, Metricbeat,
-                       Packetbeat, Logstash, Kafka, Heartbeat)
-
-from ..compose import Zookeeper
-
+from scripts.apm_agents import AgentGoNetHttp, AgentNodejsExpress, AgentPythonDjango, AgentPythonFlask, AgentRubyRails, \
```
v1v: typo?

scripts/compose.py
```diff
@@ -1,26 +1,22 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
v1v: that's the right one?

scripts/compose.py
```diff
@@ -34,16 +30,8 @@
 PACKAGE_NAME = 'localmanager'
 __version__ = "4.0.0"
 
-DEFAULT_STACK_VERSION = "8.0"
```
v1v: What's the reason for moving the version to another class?

kuisathaverat: this constant is used by services class so it was moved to the same file.

scripts/compose.py
kuisathaverat: to make it works in Python 2.7 we have to edit the `PYTHONPATH`, on Python 3 it is not needed, so because Python 2.7 is in its end of life I suggest to move forward to Python 3 -> https://pythonclock.org/

beniwohli: This needs to be documented in the `README` though, ideally with some instructions how to get Python 3 on OSX (not sure if `compose.py` even works on Windows)

scripts/helpers.py
beniwohli: @v1v `build_manifests` acts as a global cache here, so subsequent calls to `resolve_bc` with the same arguments don't have to make expensive network requests.

Since we're switching to Python 3, we could theoretically use [`functools.lru_cache`](https://docs.python.org/3/library/functools.html#functools.lru_cache) which achieves the same in a less hacky way, but let's leave that for after this refactor.

scripts/tests/localsetup_tests.py
beniwohli: @v1v the `\` is needed because the next line is a continuation of the import statement. I myself prefer to use parentheses:

    from scripts.opbeans import (
        X, Y, Z,
        A, B, C,
    )

scripts/helpers.py
v1v: Thanks :)

scripts/tests/localsetup_tests.py
v1v: got it

scripts/compose.py
v1v: got it

scripts/compose.py
kuisathaverat: it is true python3 is not in Mac OS X by default, we will have to live with python 2.7 for a time

beniwohli: Apple is about to deprecate the use of the included Python in OS X 10.15 in October, and will remove it completely in 10.16. So we need to document how to install Python anyway, might as well do it now with Python 3

https://www.macobserver.com/analysis/macos-catalina-deprecates-unix-scripting-languages/

kuisathaverat: I've made it works on Python 2.7

kuisathaverat:
@beniwohli I'd need your Python super powers with the `@mock.patch` there are 4 test that does not replace the method `modules.helpers.resolve_bc`, I have tried multiple paths and finally, I keep the original value to see if you see something that I'm not. These are the values I've tried with the associated imports to make it works

`modules.helpers.resolve_bc`
`modules.services.resolve_bc`
`compose.resolve_bc`
`helpers.__name__ + ".resolve_bc" `
`compose.__name__ + ".resolve_bc"` 
`resolve_bc.__module__ + resolve_bc.__name__`
`scripts.modules.helpers.resolve_bc` <--- on debug mode this is the path if  you inspect resolve_bc.__module__ 
`scripts.modules.services.resolve_bc`

one of the test is `LocalTest.test_docker_download_image_url` in the file `scripts/tests/localsetup_tests.py`

kuisathaverat:
I 've resolved the mocking issues, I've moved all the code to modules and keep the compose as a simple wrapper, then mocks works

kuisathaverat:
CI test errors are not related to the PR

kuisathaverat:
@v1v @graphaelli @beniwohli please take a look to this one, because it is a big refactor every PR we merge makes me to manually patch this PR, I do not want to keep it open for a long time.

scripts/__init__.py
```diff
@@ -0,0 +1,6 @@
+#from . import modules
```
beniwohli: should we remove these commented-out imports completely? I suppose they are left-overs from before you re-shuffled the structure

kuisathaverat: yep, those are leftovers, I've removed them

graphaelli:
I haven't had a chance for code-level review but the idea sounds good to me.

kuisathaverat:
@graphaelli so, you are agreed to go ahead or you want to take a deeper look

graphaelli:
Please go ahead, all yours

PR #602, opened 2019-09-20
fix: add a call to add_arguments supper method on Kibana, some test to check the missing options

### LINKED ISSUE #601: --kibana-port parameter does not work
something in https://github.com/elastic/apm-integration-testing/pull/557 removes some Kibana parameters.

kuisathaverat:
I've introduced a regression by adding a new method `add_arguments`to Kibana, basically, I've removed all configuration options because I've missed adding a call to the superclass method. It is fixed on this PR, also I've added some test to avoid this happens in the future.

closes #601

### DIFF
--- a/scripts/tests/service_tests.py
+++ b/scripts/tests/service_tests.py
@@ -885,6 +885,22 @@ class KibanaServiceTest(ServiceTest):
         self.assertTrue("elasticsearch" in kibana['depends_on'])
         self.assertEqual("elasticsearch01:9200,elasticsearch02:9200", kibana['environment']["ELASTICSEARCH_URL"])
 
+    def test_kibana_port(self):
+        kibana = Kibana(version="7.3.0", release=True, kibana_port="1234").render()["kibana"]
+        self.assertTrue("127.0.0.1:1234:5601" in kibana['ports'])
+
+    def test_kibana_version(self):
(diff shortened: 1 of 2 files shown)


scripts/tests/service_tests.py
```diff
@@ -885,6 +885,22 @@ def test_kibana_elasticsearch_urls(self):
         self.assertTrue("elasticsearch" in kibana['depends_on'])
         self.assertEqual("elasticsearch01:9200,elasticsearch02:9200", kibana['environment']["ELASTICSEARCH_URL"])
 
+    def test_kibana_port(self):
+        kibana = Kibana(version="7.3.0", release=True, kibana_port="1234").render()["kibana"]
```
mdelapenya: Is it possible to extract version to a variable? That will allow changing one place in future modifications

Thinking out loud, even the kibana object could be extracted to a global variable

kuisathaverat: in most cases, it is a value that has to match with the result of the test, if we move it to a class field we will still have to modify every assert to match the new version, or we make the result string in runtime.

kuisathaverat: BTW the version is not important for the test, I mean it could be any value does not change anything (e.g. `10.100.1` is also valid).

kuisathaverat:
Failed tests are not related to the PR

PR #606, opened 2019-09-23
fix: disable elasticdump

### LINKED ISSUE #560: Allow disable Elasticsearch dump on test failure
if you execute `make test-compose` you would need an Elasticsearch instance running locally to avoid have a timeout on every test failure dump, this makes tests take too much time. We could add an environment variable to enable disable those dumps, by default it should be disabled.

kuisathaverat:
closes #560

### DIFF
--- a/tests/agent/concurrent_requests.py
+++ b/tests/agent/concurrent_requests.py
@@ -112,7 +112,6 @@ class Concurrent:
 
         def assert_count(terms, expected):
             """wait a bit for doc count to reach expectation"""
-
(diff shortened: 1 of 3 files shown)


tests/agent/concurrent_requests.py
```diff
@@ -112,7 +112,6 @@ def check_counts(self, it, max_wait=60, backoff=1):
 
         def assert_count(terms, expected):
             """wait a bit for doc count to reach expectation"""
```
kuisathaverat: lint fix

PR #616, opened 2019-09-25
Enable JSON parsing and log correlation for opbeans-java

felixbarny:


felixbarny:
Thx @basepi for paving the way 🙂

Issue #596, opened 2019-09-19
enable codesearch


graphaelli:
per the [docs](https://www.elastic.co/guide/en/kibana/current/code-getting-started.html), `xpack.code.ui.enabled: true` when on default distribution 7.2+.

graphaelli:
depends on https://github.com/elastic/kibana/issues/46140

PR #617, opened 2019-10-02
Add support for Elastic Code from 7.2+

v1v:
## What does this PR do?

Enable Elastic Code by default when version is +7.2

## Why is it important?

Welcome Code team!! :)

## Related issues
Closes https://github.com/elastic/apm-integration-testing/issues/596

## Tests

```bash
scripts/compose.py start master
```

image


graphaelli:
I'm a fan and had started working on this but https://github.com/elastic/kibana/pull/46151 will prevent it from working in < 7.5 so merging this in may be misleading.

I think code will be on by default in 7.5+ so there may be no need for this as is.

v1v:
I'll then close this PR and wait for the 7.5 release  Thanks Gil for the heads up, and I actually worked on this since I did work on supporting the same for the K8s clusters so I realized this issue was opened too.

PR #620, opened 2019-10-07
Bump dotnet agent version 1.1.1

v1v:
## What does this PR do?

Bump version [1.1.1](https://github.com/elastic/apm-agent-dotnet/releases/tag/1.1.1)

## Why is it important?

Keep up to date

## Related issues
Closes #ISSUE

### DIFF
--- a/docker/dotnet/aspnetcore/TestAspNetCoreApp.csproj
+++ b/docker/dotnet/aspnetcore/TestAspNetCoreApp.csproj
@@ -6,7 +6,7 @@
     </PropertyGroup>
 
     <ItemGroup>
-        <PackageReference Include="Elastic.Apm.NetCoreAll" Version="1.0.1" />
+        <PackageReference Include="Elastic.Apm.NetCoreAll" Version="1.1.1" />
         <PackageReference Include="Microsoft.AspNetCore.App" />
         <PackageReference Include="Microsoft.AspNetCore.Razor.Design" Version="2.2.0" PrivateAssets="All" />
     </ItemGroup>


mdelapenya:
Is this task something that could be automated on every .NET release?

gregkalapos:
> Is this task something that could be automated on every .NET release?

That's a good question! 

NuGet supports [floating versions](https://docs.microsoft.com/en-us/nuget/concepts/dependency-resolution#floating-versions) - so we could add:

```
<PackageReference Include="Elastic.Apm.NetCoreAll" Version="1.*" />
```

That would mean: always fetch the latest 1.x version, whatever that is. 

We could also do:

```
<PackageReference Include="Elastic.Apm.NetCoreAll" Version="*" />
```

which is I guess: fetch the latest - but maybe with this we'd go too far... at least on majors  I think we should manually pump the version, mainly because those can have breaking changes, so the build would potentially be broken anyway, because we need to adapt the C# code also in this project.

mdelapenya:
I'd prefer using fixed versions, as it explicitly tells us the version the-project-depending-on-the-dependency works.

gregkalapos:
> I'd prefer using fixed versions, as it explicitly tells us the version the-project-depending-on-the-dependency works.

If that helps, I think we can figure out the selected version fairly easily - E.g. we can add some code into the sample app that prints it. I played with the `dotnet` CLI tool that fetches the package - I did not find an easy way to print the selected version with that, but maybe there is a way to do it even with that - That would mean just add some parameter to the `dotnet` command and we'd see the selected agent version in the logs.

If we want to manually fetch the latest agent version and modify the `.csproj` file then I think that'd be a little bit more effort - you (the automation team) know probably more about options for that route than I do - but I'm happy to help out.

v1v:
The current test failures are not related to this particular PR but already something reported to the rum agent

PR #626, opened 2019-10-11
bump 1.1.2 dotnet

v1v:


v1v:
We tracked this particular automation use case to investigate how to proceed, meanwhile, we might need to keep doing this PRs manually 😿

PR #628, opened 2019-10-16
fix: upgrade test

kuisathaverat:
## What does this PR do?

fix the docker container used for the upgrade test, also it uses a release for the stack

## Why is it important?

the test is broken.

## Related issues
Closes https://github.com/elastic/observability-dev/issues/395

### DIFF
--- a/docker/go/nethttp/Dockerfile
+++ b/docker/go/nethttp/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:latest
+FROM golang:1.8
 ENV GO111MODULE=on
 WORKDIR /src/testapp
 COPY . /src/testapp
--- a/.ci/scripts/7.0-upgrade.sh
+++ b/.ci/scripts/7.0-upgrade.sh
@@ -2,14 +2,15 @@
 
 srcdir=$(dirname "$0")
 test -z "$srcdir" && srcdir=.
+# shellcheck source=/dev/null
 . "${srcdir}/common.sh"
 
-export COMPOSE_ARGS="6.6 --force-build --build-parallel \
+export COMPOSE_ARGS="6.6 --release --force-build --build-parallel \
   --no-apm-server-dashboards --no-apm-server-self-instrument \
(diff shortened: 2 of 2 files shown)


docker/go/nethttp/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:latest
+FROM golang:1.8
```
mdelapenya: Why Go 1.8 and no any other?

kuisathaverat: it is the minimum version required, it is defined in the Agent Go README

axw:
@kuisathaverat why did you change it to use Go 1.8?

This is causing errors: https://apm-ci.elastic.co///blue/rest/organizations/jenkins/pipelines/apm-integration-test-downstream/pipelines/7.x/runs/948/steps/58/log/

The reason is that Go 1.8 does not support Go Modules, so the go.mod file is ignored and the test program doesn't know where to find the Go agent code.

mdelapenya:
Sorry for our change, let's fix this. I guess we need Go 1.11 as minimum version, but just to use same version for development than for testing, which version is the correct one?

Maybe we should update this too: https://github.com/elastic/apm-agent-go#requirements

axw:
@mdelapenya no worries, I was just wondering why it was changed - I figured there must be a reason I just couldn't see it.

> I guess we need Go 1.11 as minimum version, but just to use same version for development than for testing, which version is the correct one?

I don't use a specific version of Go in development. Unit tests are run using Go 1.12.x. I think we should be safe to use any version from 1.11.x onward, as we have coverage of all supported Go versions in the unit tests.

> Maybe we should update this too: https://github.com/elastic/apm-agent-go#requirements

The agent supports Go 1.8, it's just the way to integration tests work that precludes running with that version. I plan to deprecate support for Go 1.8 in the not-too-distant future, though, so I don't think it's worth trying to change the integration tests to support it. We have good enough coverage within the Go agent repo itself.

PR #629, opened 2019-10-16
Fix upgrade test

kuisathaverat:
## What does this PR do?

it changes to build a fixed version of the APM agents because compose does not support to pass the version to Opbeans, the script builds the images manually. Fix the Opbeans Go that does not start due to https://github.com/GoogleContainerTools/distroless/issues/390

## Why is it important?

The test is still broken.

## Related issues
Closes https://github.com/elastic/observability-dev/issues/395

### DIFF
--- a/.ci/scripts/7.0-upgrade.sh
+++ b/.ci/scripts/7.0-upgrade.sh
@@ -2,14 +2,15 @@
 
 srcdir=$(dirname "$0")
 test -z "$srcdir" && srcdir=.
+# shellcheck source=/dev/null
 . "${srcdir}/common.sh"
 
-export COMPOSE_ARGS="6.6 --force-build --build-parallel \
+export COMPOSE_ARGS="6.6 --release --force-build --build-parallel \
   --no-apm-server-dashboards --no-apm-server-self-instrument \
-  --apm-server-count 2 --apm-server-tee --elasticsearch-data-dir '' \
+  --elasticsearch-data-dir '' \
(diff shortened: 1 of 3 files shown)


docker/opbeans/go/Dockerfile
```diff
@@ -19,7 +20,7 @@ RUN go build
 
 # Stage 1: copy static assets from opbeans/opbeans-frontend and
 # opbeans-go from stage 0 into a minimal image.
-FROM gcr.io/distroless/base
+FROM gcr.io/distroless/base-debian10
```
mdelapenya: How big is the debian image related to the previous one?

.ci/scripts/7.0-upgrade.sh
```diff
@@ -5,16 +5,19 @@ test -z "$srcdir" && srcdir=.
 # shellcheck source=/dev/null
 . "${srcdir}/common.sh"
 
-export COMPOSE_ARGS="6.6 --release --force-build --build-parallel \
+docker build --build-arg GO_AGENT_BRANCH=v1.5.0 -t localtesting_6.6.2_opbeans-go docker/opbeans/go
```
v1v: `--opbeans-go-agent-branch=1.x\` was replaced so, if this required too?

kuisathaverat: really we want to use the current version, we tried with the branch but the branch would change in a few for sure so we've moved a fixed version that works, the 1.5.0 works.,

Issue #636, opened 2019-10-17
filebeat report errors in the log


kuisathaverat:
`./scripts/compose.py start master --with-opbeans-java --with-filebeat`

```
{"level":"error","timestamp":"2019-10-17T17:41:37.235Z","caller":"readjson/json.go:52","message":"Error decoding JSON: json: cannot unmarshal number into Go value of type map[string]interface {}"}
{"level":"error","timestamp":"2019-10-17T17:41:38.852Z","logger":"autodiscover","caller":"cfgfile/list.go:96","message":"Error creating runner from config: Can only start an input when all related states are finished: {Id:2395809-2049 Finished:false Fileinfo:0xc0007561a0 Source:/var/lib/docker/containers/d9da67140c825712499ecc333458d990d0b1082702a954761683792d21782bbd/d9da67140c825712499ecc333458d990d0b1082702a954761683792d21782bbd-json.log Offset:609552 Timestamp:2019-10-17 17:41:36.79576 +0000 UTC m=+256.293930701 TTL:-1ns Type:docker Meta:map[] FileStateOS:2395809-2049}"}
```

kuisathaverat:
there are some log messages that are not JSON 

```
{"log":"2019-10-17 17:55:04.803 [apm-remote-config-poller] INFO co.elastic.apm.agent.configuration.ApmServerConfigurationSource - Received new configuration from APM Server: {}\n","stream":"stdout","time":"2019-10-17T17:55:04.8041885Z"}
```

The configuration should ignore them but I am not sure

```
        - condition:
            contains:
              docker.container.image: "opbeans-java"
          config:
            - type: docker
              containers.ids:
                - "${data.docker.container.id}"
              processors:
                - if:
                    regexp:
                      message: '^\{.*\}$'
                  then:
                    - decode_json_fields:
                        fields:
                          - message
                        target: ""
                        overwrite_keys: true
              fields_under_root: true
```

kuisathaverat:
It is reading the logs of all containers on every container config, I think it is related to the path setting `- '/var/lib/docker/containers/*/*.log'`


```
        - condition:
            contains:
              docker.container.image: "opbeans-java"
          config:
            - type: container
              paths:
                - '/var/lib/docker/containers/*/*.log'
              containers.ids:
                - "${data.docker.container.id}"
              tail_files: true
              tags: ["opbeans-java"]
```

[...]

PR #637, opened 2019-10-18
fix: Filebeat configuration

kuisathaverat:
## What does this PR do?

Set a tested Filebeat configuration for each container deploy by the compose.py

## Why is it important?

The current configuration parses all log container per input configuration thus make that all log lines from all containers are indexed for every container, also, it is trying to parse some lines that are not in the expected format so Filebeat has a ton of errors in logs, and those lines are not properly indexed.

## Related issues
Closes https://github.com/elastic/apm-integration-testing/issues/636 #639


docker/filebeat/filebeat.yml
```diff
@@ -25,150 +25,201 @@ filebeat.autodiscover:
...
+                - "/var/lib/docker/containers/*/${data.docker.container.id}-json.log"
               containers.ids:
                 - "${data.docker.container.id}"
+              tail_files: true
               fields_under_root: true
               json.keys_under_root: true
               json.overwrite_keys: true
               json.add_error_key: true
               json.message_key: message
+              include_lines: ["^{"]
```
beniwohli: Doesn't this simply drop log lines that aren't JSON?

We have a similar issue with opbeans-python, and there we (as in, @basepi) went with a conditional:

https://github.com/elastic/apm-integration-testing/blob/971186146810d008794a158240f5329954460287/docker/filebeat/filebeat.yml#L129-L134

(I see you removed that conditional as well in this PR, is that intentional?)

kuisathaverat: define this processor is the same that include only JSON lines, all the rest of lines are ignored because there is no alternative definition to process them.

>(I see you removed that conditional as well in this PR, is that intentional?)

yes, I did no see any issues with the Python log so I've removed it, we can include also the include setting to avoid issues.

beniwohli: When using the current master, I see a lot of log lines that don't match that regex in the processor, e.g.

    [2019-10-21 11:55:27,123: INFO/MainProcess] Received task: opbeans.tasks.update_stats[bdadda02-95a8-4f93-a1c5-47a6b93f52ae] 

So it doesn't look to me as if they are getting dropped.

I don't think we should drop _any_ log lines. While it's not really mission critical for ourselves, users will see this filebeat.yml in google searches etc. and use it as starting point for their own setups. We should avoid potential footguns like "drop anything that doesn't look like JSON".

kuisathaverat: agree, but with your filter, we lose any JSON field that is not in the message field and the original message field

i.e. date and level will be lost

```
{ date: "...", level: "...", message: "..."}
```

I am going to rethink the thing to try to keep it as simple as possible without lose anything, we do not want to maintain 10 different ways to process the logs

kuisathaverat: I've realized that the Python Agent does not have a pure JSON output this is the most similar message toa JSON, you should preprocess it by removing the first part of the message before to parse the JSON

```
[2019-10-21 16:22:12,044: DEBUG/MainProcess] {'event': 'flushing due to time since last flush 10.544s > max_flush_time 10.514s', '@timestamp': '2019-10-21T16:22:12.044753Z', 'logger': 'elasticapm.transport', 'level': 'debug'}
```

beniwohli: Yes, we're aware of that, that's why we introduced that processor I linked to above. Unfortunately, it's not an easy fix, as this logging format is set up by Celery (the background task manager). We spent some time in trying to fix it, but didn't find a good solution yet. Hence we'd really prefer if the processor could be kept in place for now.

kuisathaverat: I was wrong there are some JSON lines in Python logs

kuisathaverat:
finally, we found a common configuration for all Opbeans, the fields in the message are stored in a new field named `json` to avoid some issues replacing fields on the root of the object, also in that way the original message is keep

```
localtesting_7.4.0_apm-server | {"level":"info","timestamp":"2019-10-22T07:58:45.515Z","logger":"request","caller":"middleware/log_middleware.go:76","message":"request ok","request_id":"da00161e-0352-49fc-90d1-247d4becf037","method":"POST","URL":"/config/v1/agents","content_length":68,"remote_address":"172.19.0.8","user-agent":"elasticapm-python/5.2.2","response_code":200}
```

beniwohli:
@kuisathaverat will that still work with log correlation? The APM/Logs UI expect the trace id in the `trace.id` field, I don't think it'll work with `json.trace.id`. Or I might misunderstand your solution.

kuisathaverat:
>@kuisathaverat will that still work with log correlation? The APM/Logs UI expects the trace id in the trace.id field, I don't think it'll work with json.trace.id. Or I might misunderstand your solution.

In that case, we will find a solution to overwrite the timestamp field

kuisathaverat:
@beniwohli Python is generating log lines like this one `{"event": "forced flush", "@timestamp": "2019-10-22T09:50:41.819139Z", "logger": "elasticapm.transport", "level": "debug"}{"event": "Internal Server Error: /api/products/top"}` it contains two JSON records in the same line this makes imposible to parse it.

kuisathaverat:
puff, we will hit mapping issues between Opbeans fields types

```
localtesting_7.4.0_filebeat | {"level":"warn","timestamp":"2019-10-22T10:00:50.993Z","caller":"elasticsearch/client.go:535","message":"Cannot index event publisher.Event{Content:beat.Event{Timestamp:time.Time{wall:0x1f54438, ext:63707335248, loc:(*time.Location)(nil)}, Meta:common.MapStr(nil), Fields:common.MapStr{\"agent\":common.MapStr{\"ephemeral_id\":\"79dcb0af-29b2-4173-a1ae-247f0fb02500\", \"hostname\":\"cbd336f9f92d\", \"id\":\"f2bbd3ed-ca93-4307-8936-b55421590b96\", \"type\":\"filebeat\", \"version\":\"7.4.0\"}, \"container\":common.MapStr{\"id\":\"6cbf37f60dce644ca2038be8f4644a75cdf2973ef0bd8a24ee529cd823a16406\", \"image\":common.MapStr{\"name\":\"apm-integration-testing_opbeans-python\"}, \"name\":\"localtesting_latest_opbeans-python\"}, \"docker\":common.MapStr{\"container\":common.MapStr{\"labels\":common.MapStr{\"com_docker_compose_config-hash\":\"0ed229cafd4459810b2d4941fb7e84a393b347d0595a732f4eabaced455030c0\", \"com_docker_compose_container-number\":\"1\", \"com_docker_compose_oneoff\":\"False\", \"com_docker_compose_project\":\"apm-integration-testing\", \"com_docker_compose_service\":\"opbeans-python\", \"com_docker_compose_version\":\"1.24.1\"}}}, \"ecs\":common.MapStr{\"version\":\"1.1.0\"}, \"event\":\"dt_ping_pong_connection_failed\", \"host\":common.MapStr{\"name\":\"cbd336f9f92d\"}, \"input\":common.MapStr{\"type\":\"container\"}, \"level\":\"error\", \"log\":common.MapStr{\"file\":common.MapStr{\"path\":\"/var/lib/docker/containers/6cbf37f60dce644ca2038be8f4644a75cdf2973ef0bd8a24ee529cd823a16406/6cbf37f60dce644ca2038be8f4644a75cdf2973ef0bd8a24ee529cd823a16406-json.log\"}, \"offset\":569505}, \"logger\":\"opbeans.views\", \"message\":\"{\\\"service\\\": \\\"http://opbeans-dotnet:3000\\\", \\\"event\\\": \\\"dt_ping_pong_connection_failed\\\", \\\"@timestamp\\\": \\\"2019-10-22T10:00:48.032851Z\\\", \\\"logger\\\": \\\"opbeans.views\\\", \\\"level\\\": \\\"error\\\", \\\"transaction.id\\\": \\\"235450ec5419b5ce\\\", \\\"trace.id\\\": \\\"58b5f7f0d572777dab18af019d0430e5\\\"}\", \"parser_type\":[]string{\"json\"}, \"service\":\"http://opbeans-dotnet:3000\", \"stream\":\"stdout\", \"trace.id\":\"58b5f7f0d572777dab18af019d0430e5\", \"transaction.id\":\"235450ec5419b5ce\"}, Private:file.State{Id:\"\", Finished:false, Fileinfo:(*os.fileStat)(0xc000833040), Source:\"/var/lib/docker/containers/6cbf37f60dce644ca2038be8f4644a75cdf2973ef0bd8a24ee529cd823a16406/6cbf37f60dce644ca2038be8f4644a75cdf2973ef0bd8a24ee529cd823a16406-json.log\", Offset:569862, Timestamp:time.Time{wall:0xbf63d27f0c63b760, ext:59597810301, loc:(*time.Location)(0x4de4580)}, TTL:-1, Type:\"container\", Meta:map[string]string(nil), FileStateOS:file.StateOS{Inode:0x26e090, Device:0x801}}, TimeSeries:false}, Flags:0x1} (status=400): {\"type\":\"mapper_parsing_exception\",\"reason\":\"object mapping for [service] tried to parse field [service] as object, but found a concrete value\"}"}
```

Also, the same field has different names on each Opbean, e.i `msg`, `message`, and `event` are the different names for the log message, it depends of the Opbean.

kuisathaverat:
I've talked with @ruflin this morning, I hope he can help us 🙏 but I start thinking that before of anything we have to normalize the name and the type of the fields on every Opbeans project.

docker/filebeat/filebeat.yml
```diff
@@ -25,190 +25,142 @@ filebeat.autodiscover:
       templates:
         - condition:
             contains:
-              docker.container.image: "apm-server"
+              docker.container.name: "opbeans-"
```
ruflin: How many different `opbeans-*` container types do we have? As the logs coming from these containers are so different, I wonder if we should treat them differently.

docker/filebeat/filebeat.yml
```diff
@@ -25,190 +25,142 @@ filebeat.autodiscover:
...
+                  add_error_key: true
               fields_under_root: true
-              json.keys_under_root: true
-              json.overwrite_keys: true
-              json.add_error_key: true
-              json.message_key: message
         - condition:
             contains:
-              docker.container.image: "filebeat"
+              docker.container.name: "opbeans-"
```
ruflin: Same condition twice?

docker/filebeat/filebeat.yml
kuisathaverat: >How many different opbeans-* container types do we have? 

There are 7 different agents (.NET, Java, Go, Node.js, Python, Ruby, and RUM)

>As the logs coming from these containers are so different

We'd try that they aren't, but right now they are different

>I wonder if we should treat them differently.

Do you mean different configs or different indices?

docker/filebeat/filebeat.yml
kuisathaverat: yep, this is only for the no JSON log lines, I know that we can process it in the same config, but for the moment to make both config as simple as possible I've split the config in two, JSON and NO_JSON

docker/filebeat/filebeat.yml
ruflin: With the FB setup today, it will make your life much easier if we can get them into index. Because of this I'm thinking of different configs like you already did for json and non json to split it up even further. Like this if 5-6 work, you can move the data of the conflicting one for example under its own prefix.

docker/filebeat/filebeat.yml
```diff
@@ -25,190 +25,142 @@ filebeat.autodiscover:
...
+              multiline.match: after
+              exclude_lines: ['^{']
+              processors:
+              - add_tags:
+                  tags: [no_json]
+                  target: "parser_type"
         - condition:
             contains:
-              docker.container.image: "heartbeat"
+                docker.container.name: "kibana"
```
v1v: Not sure whether this is a cosmetic ident or it might imply something else
```suggestion
              docker.container.name: "kibana"
```

docker/filebeat/filebeat.yml
```diff
@@ -25,190 +25,142 @@ filebeat.autodiscover:
...
+                - "/var/lib/docker/containers/*/${data.docker.container.id}-json.log"
+              tail_files: true
               json.add_error_key: true
-              json.message_key: message
+              json.overwrite_keys: true
+              json.keys_under_root: true
         - condition:
             contains:
-              docker.container.image: "kibana"
+                docker.container.name: "elasticsearch"
```
v1v: ditto
```suggestion
              docker.container.name: "elasticsearch"
```

docker/filebeat/filebeat.yml
```diff
@@ -25,190 +25,142 @@ filebeat.autodiscover:
...
+                - "/var/lib/docker/containers/*/${data.docker.container.id}-json.log"
+              tail_files: true
               json.add_error_key: true
-              json.message_key: message
+              json.overwrite_keys: true
+              json.keys_under_root: true
         - condition:
             contains:
-              docker.container.image: "metricbeat"
+                docker.container.name: "metricbeat"
```
v1v: ditto
```suggestion
              docker.container.name: "metricbeat"
```

docker/filebeat/filebeat.yml
```diff
@@ -25,190 +25,142 @@ filebeat.autodiscover:
...
+                - "/var/lib/docker/containers/*/${data.docker.container.id}-json.log"
+              tail_files: true
               json.add_error_key: true
-              json.message_key: message
+              json.overwrite_keys: true
+              json.keys_under_root: true
         - condition:
             contains:
-              docker.container.image: "opbeans-node"
+                docker.container.name: "heartbeat"
```
v1v:
```suggestion
              docker.container.name: "heartbeat"
```

docker/filebeat/filebeat.yml
```diff
@@ -25,190 +25,142 @@ filebeat.autodiscover:
...
-              multiline.match: after
+                - "/var/lib/docker/containers/*/${data.docker.container.id}-json.log"
+              tail_files: true
+              json.add_error_key: true
+              json.overwrite_keys: true
+              json.keys_under_root: true
         - condition:
             contains:
-              docker.container.image: "opbeans-go"
+                docker.container.name: "filebeat"
```
v1v:
```suggestion
              docker.container.name: "filebeat"
```

docker/filebeat/filebeat.yml
```diff
@@ -25,190 +25,142 @@ filebeat.autodiscover:
...
-                        fields:
-                          - from: "event"
-                            to: "message"
-              fields_under_root: true
+              json.overwrite_keys: true
+              json.keys_under_root: true
         - condition:
             contains:
-              docker.container.image: "opbeans-java"
+                docker.container.name: "apm-server"
```
v1v:
```suggestion
              docker.container.name: "apm-server"
```

kuisathaverat:
depends on https://github.com/elastic/opbeans-ruby/pull/23

kuisathaverat:
this PR is ready to merge, after merge it Java, Go, Python, and Ruby would have log correlation

PR #641, opened 2019-10-21
Adds a basic agent remote config test.

### LINKED ISSUE #436: Add automated test for agents CM
At minimum we need one agent to test that the whole feature (Kibana-server-agent), but ideally  we add tests for all the agents progressively to ensure they can pick up and apply configuration dynamically.

jalvz:
Fixes #436

### DIFF
--- a/scripts/tests/localsetup_tests.py
+++ b/scripts/tests/localsetup_tests.py
@@ -818,7 +818,7 @@ class LocalTest(unittest.TestCase):
                     -E, 'setup.kibana.host=kibana:5601', -E, setup.template.settings.index.number_of_replicas=0,
                     -E, setup.template.settings.index.number_of_shards=1, -E, setup.template.settings.index.refresh_interval=1ms,
                     -E, monitoring.elasticsearch=true, -E, monitoring.enabled=true,
-                    -E, apm-server.kibana.enabled=true, -E, apm-server.agent.config.cache.expiration=2s, -E, 'apm-server.kibana.host=kibana:5601',
+                    -E, apm-server.kibana.enabled=true, -E, apm-server.agent.config.cache.expiration=1s, -E, 'apm-server.kibana.host=kibana:5601',
                     -E, 'output.elasticsearch.hosts=["elasticsearch:9200"]', -E, output.elasticsearch.enabled=true,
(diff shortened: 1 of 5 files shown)


jalvz:
jenkins retest this please

tests/agent/__init__.py
```diff
@@ -0,0 +1,48 @@
+from contextlib import contextmanager
+import requests
+import time
+from urlparse import urljoin
+
+from tests.fixtures import default
+
+
+@contextmanager
+def remote_config(url, sampling_rate=1.0):
```
graphaelli: since `url` is used a lot I think it should be more descriptive, how about `kibana_url` ?

scripts/tests/localsetup_tests.py
```diff
@@ -818,7 +818,7 @@ def test_start_master_default(self):
                     -E, 'setup.kibana.host=kibana:5601', -E, setup.template.settings.index.number_of_replicas=0,
                     -E, setup.template.settings.index.number_of_shards=1, -E, setup.template.settings.index.refresh_interval=1ms,
                     -E, monitoring.elasticsearch=true, -E, monitoring.enabled=true,
-                    -E, apm-server.kibana.enabled=true, -E, 'apm-server.kibana.host=kibana:5601',
+                    -E, apm-server.kibana.enabled=true, -E, apm-server.agent.config.cache.expiration=2s, -E, 'apm-server.kibana.host=kibana:5601',
```
graphaelli: any reason the expiry can't be even shorter?

tests/agent/__init__.py
```diff
@@ -0,0 +1,48 @@
+from contextlib import contextmanager
+import requests
+import time
+from urlparse import urljoin
```
kuisathaverat:
```suggestion
from urllib.parse import urljoin
```

kuisathaverat: see https://stackoverflow.com/questions/48072619/how-can-i-import-urlparse-in-python-3?answertab=votes#tab-top

scripts/tests/localsetup_tests.py
jalvz: im fine with 1s, since this is going to be used for opbeans and everything else I didn't want it to be too noisy

Issue #173, opened 2018-09-11
disable opbeans loadgen when no services need it


graphaelli:
`compose.py --with-opbeans-python --no-opbeans-python-loadgen` should realize that no services need load generation and prevent addition of the service.  Currently, we end up with a service with empty `OPBEANS_URLS` instead.

```
"opbeans-load-generator": {
  "container_name": "localtesting_7.0.0-alpha1_opbeans-load-generator",
  "depends_on": {},
  "environment": [
    "OPBEANS_URLS=",
    "OPBEANS_RPMS="
  ],
  "image": "opbeans/opbeans-loadgen:latest",
  "logging": {
    "driver": "json-file",
    "options": {
      "max-file": "5",
      "max-size": "2m"
    }
  }
},
```

Issue #213, opened 2018-10-30
Option to not starting opbeans load generator


watson:
If you start any of the opbeans images, the load generator will start as well. It would be nice if there was an option to run opbeans without the load generator

watson:
The main reason why this would be nice, is that it adds about 1 minute to the boot up time to spin up the load generator on my mac even if I don't need it 😅

kuisathaverat:
related to https://github.com/elastic/apm-integration-testing/issues/173

kuisathaverat:
resolved in https://github.com/elastic/apm-integration-testing/pull/648

PR #648, opened 2019-10-23
feat: allow to disable loadgen

kuisathaverat:
## What does this PR do?

adds a new parameter `--no-opbeans-load-generator` to disable the load generator.

## Why is it important?

if you run `--all` or `--al-beans`you would have to use the `--no-opbeans-*-loadgen` parameters for every Opbean.

## Related issues
Closes https://github.com/elastic/apm-integration-testing/issues/173 https://github.com/elastic/apm-integration-testing/issues/213

### DIFF
--- a/scripts/modules/cli.py
+++ b/scripts/modules/cli.py
@@ -204,6 +204,13 @@ class LocalSetup(object):
                 )
             service.add_arguments(parser)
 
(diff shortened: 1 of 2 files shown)


mdelapenya:
Let's collect a dollar/euro/pound... per each flag in this CLI 💰🤑

Issue #578, opened 2019-08-26
Integration tests for secure communication


axw:
We should have integration tests for secure communication between APM Server and agents. At least:

 - set up a self-signed cert on the server, and ensure that agents will fail to verify the cert by default
 - set `ELASTIC_APM_VERIFY_SERVER_CERT=false` on the agents, and check that they can communicate
 - set `ELASTIC_APM_SERVER_CERT=<path/to/cert>` on the agents, and check that they can communicate

kuisathaverat:
related to https://github.com/elastic/apm-integration-testing/issues/221

Issue #221, opened 2018-11-06
add secret token to apm-server endpoint configuration


graphaelli:
success means running `tests/server/test_transaction.py` against an apm-server with a secret token set.

kuisathaverat:
related to https://github.com/elastic/apm-integration-testing/issues/13

kuisathaverat:
I think is fully covered on https://github.com/elastic/apm-integration-testing/pull/737
WDYT? @graphaelli

graphaelli:
agreed, thanks!

Issue #13, opened 2017-11-13
Add agent tests for different apm-server config.


simitt:
Add tests for using TLS and secret token.

PR #656, opened 2019-10-24
feat: add support to configure TLS on the APM Server

kuisathaverat:
## What does this PR do?

Allow configuring TLS on the APM Server

## Why is it important?

We need it to be able to make tests with TLS. After merging it we will have a couple of options to enable TLS on the APM Server and another to disable certificates verification in the agents/Opbeans. We add a couple of options to allow make tests that verify if the agent checks the certificates (see https://github.com/elastic/apm-integration-testing/issues/578)

`/scripts/compose.py start 7.4 --release --apm-server-enable-tls --with-opbeans-python --no-verify-server-cert`

## Related issues
related to https://github.com/elastic/apm-integration-testing/issues/13 https://github.com/elastic/apm-integration-testing/issues/221 and https://github.com/elastic/apm-integration-testing/issues/578


Makefile
```diff
@@ -36,6 +36,10 @@ venv: requirements.txt
 lint: venv
 	flake8 --ignore=D100,D101,D102,D103,D104,D105,D106,D107,D200,D205,D400,D401,D403,W504  tests/ scripts/compose.py scripts/modules
 
+.PHONY: create-x509-cert
+create-x509-cert:
+	openssl req -x509 -newkey rsa:4096 -keyout scripts/tls/key.pem -out scripts/tls/cert.crt -days 365 -subj '/CN=test-certificate' -nodes
```
axw: I think the common name should be "apm-server"? Otherwise the agents will not be able to verify the hostname.

simitt:
LGTM, thanks for adding this. 

Related to @axw 's comment - there is some [logic in apm-server tests](https://github.com/elastic/apm-server/blob/master/tests/system/config/create_certs.sh) creating a CA and signing the certificate with it, maybe you can reuse some of that here.

Makefile
kuisathaverat: SNI is not verified if you disable the certificates verification, but yep we can change it to `apm-server`

kuisathaverat:
>Related to @axw 's comment - there is some logic in apm-server tests creating a CA and signing the certificate with it, maybe you can reuse some of that here.

it is more or less the same a selfsigned certificate, but in the APM logic is signed by a selfsigned CA, it is more complex, the result is the same a certificate that you have to trust directly on it, unless you have a strong reason to have a CA, we would keep it simple.

kuisathaverat:
>Do we need a limited year certificate or could generate another one with a more long expiration?

we can generate a 10 years certificate and forget about it, also the command is in a make file dso you can generate a new one if you want to change it.

PR #660, opened 2019-10-30
fix: add kibana to tests launched on PRs

kuisathaverat:
## What does this PR do?

remove the `--no-kibana` flags on tests launched for PRs.

## Why is it important?

It is need to test the global configuration.


kuisathaverat:
@beniwohli I guess so, in any case, start Kibana Docker container is not a problem, it would add less than a minute to the build and a few resources that we have.

PR #663, opened 2019-11-12
Cache docker images in the workers

v1v:
## What does this PR do?

Cache the missing docker images which are used for running the apm-integration-testing, therefore the build might reduce just a bit for running the same docker pull from dockerhub

## Why is it important?

Let's waste of time and therefore faster builds.

## Related issues
Closes #ISSUE


.ci/packer_cache.sh
```diff
@@ -28,26 +31,24 @@ haproxy:1.9
...
 node:12.0
 node:13
 node:13.0
 node:14
 node:14.0
-node:6
-node:6.0
-node:8
 node:8-slim
-node:8.1
```
kuisathaverat: if you remove this one here probably we have to add it to the Node.js repo

v1v: `8.1` is not an active version anymore:
- https://github.com/elastic/apm-agent-nodejs/blob/master/.ci/.jenkins_nodejs.yml

Caused by:
- https://github.com/elastic/apm-agent-nodejs/commit/408bffdeb4e44ad26f5f4086720ba725ebf19db8#diff-fe5284f01f6313792b5e5defc5b319c2

PR #664, opened 2019-11-12
timeouts resistance for the loadgen in the opbeans-rum project

### LINKED ISSUE #611: Fix the timeout issue with the opbeans RUM load generator
Currently if the page load takes longer than 10000 it throws an error and exists the process:

```
Error: wait(page load event) timed out after 10000ms
    at Timeout._onTimeout (/home/chrome/node_modules/chromeless/dist/src/util.js:180:32)
    at ontimeout (timers.js:498:11)
    at tryOnTimeout (timers.js:323:5)
    at Timer.listOnTimeout (timers.js:290:5)
```

We should either increase the timeout or catch and handle this error properly!

v1v:
## What does this PR do?

Quick fix the timeouts in the loadgen

## Why is it important?

Ensure no environmental issues when the load of the page is a bit slow

## Related issues
Closes elastic/apm-integration-testing#611

### DIFF
--- a/docker/opbeans/rum/tasks.js
+++ b/docker/opbeans/rum/tasks.js
@@ -10,34 +10,30 @@ function sleep(ms) {
 }
 
 async function run() {
-    try {
-        const chromeless = new Chromeless({
-            lauchChrome: true
-        })
+    const chromeless = new Chromeless({
+        lauchChrome: true
+    })
 
-        for (; ;) {
-            url = await chromeless
-                .goto(url)
-                .evaluate((baseUrl) => {
-                    // activateLoadGeneration is defined in opbeans-frontend
-                    if (typeof window.activateLoadGeneration === 'function') {
-                        console.log('Activating route change load generation')
-                        window.activateLoadGeneration()
-                    }
+    for (; ;) {
+        url = await chromeless
+            .goto(url)
+            .evaluate((baseUrl) => {
+                // activateLoadGeneration is defined in opbeans-frontend
+                if (typeof window.activateLoadGeneration === 'function') {
+                    console.log('Activating route change load generation')
+                    window.activateLoadGeneration()
+                }
 
-                    var links = document.querySelectorAll('a[href^="/"]')
-                    if (links && links.length) {
-                        var i = Math.floor(Math.random() * links.length)
-                        return links[i].href
-                    } else {
-                        return baseUrl
-                    }
-                }, baseUrl);
-            console.log(url)
-            await sleep(8000 + Math.floor(Math.random() * 10000))
-        }
-    } catch (e){
-      run();
+                var links = document.querySelectorAll('a[href^="/"]')
+                if (links && links.length) {
+                    var i = Math.floor(Math.random() * links.length)
+                    return links[i].href
+                } else {
+                    return baseUrl
+                }
+            }, baseUrl);
+        console.log(url)
+        await sleep(8000 + Math.floor(Math.random() * 10000))
     }
 }
 
--- a/docker/opbeans/rum/tasks.js
+++ b/docker/opbeans/rum/tasks.js
@@ -10,30 +10,34 @@ function sleep(ms) {
 }
 
 async function run() {
-    const chromeless = new Chromeless({
-        lauchChrome: true
-    })
+    try {
+        const chromeless = new Chromeless({
+            lauchChrome: true
+        })
 
-    for (; ;) {
-        url = await chromeless
-            .goto(url)
-            .evaluate((baseUrl) => {
-                // activateLoadGeneration is defined in opbeans-frontend
-                if (typeof window.activateLoadGeneration === 'function') {
-                    console.log('Activating route change load generation')
-                    window.activateLoadGeneration()
-                }
+        for (; ;) {
+            url = await chromeless
+                .goto(url)
+                .evaluate((baseUrl) => {
+                    // activateLoadGeneration is defined in opbeans-frontend
+                    if (typeof window.activateLoadGeneration === 'function') {
+                        console.log('Activating route change load generation')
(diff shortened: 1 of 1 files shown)


docker/opbeans/rum/tasks.js
```diff
@@ -10,30 +10,34 @@ function sleep(ms) {
...
+                        var i = Math.floor(Math.random() * links.length)
+                        return links[i].href
+                    } else {
+                        return baseUrl
+                    }
+                }, baseUrl);
+            console.log(url)
+            await sleep(8000 + Math.floor(Math.random() * 10000))
+        }
+    } catch (e){
```
kuisathaverat: if the RUM agent is nor running for some reason, it would be stuck forever

kuisathaverat: or when the stack is full

v1v: If that's the case the timeout in the pipeline should kill it, not ideal, neither I know whether the use case of running the opbeans-rum without the RUM agent might be something which might happen.

docker/opbeans/rum/tasks.js
```diff
@@ -10,30 +10,34 @@ function sleep(ms) {
...
+      run();
```
mdelapenya: This is a recursion, right? If so, I'd recommend having a base case with an exit condition to avoid infinite loops.

I know that the pipeline will kill it, but just to make the code resilient and decoupled from the build tool

v1v:
I didn't put enough context about the reason for using this approach, it's just a quick fix as discussed with the RUM team. As it seems a bit controversial, I'll hold this PR and work on using a kind of controlled loop to iterate a few times to avoid any kind of unexpected recurrency.

docker/opbeans/rum/tasks.js
kuisathaverat: I bet the time out is related to this line https://github.com/elastic/opbeans-frontend/blob/980651d5ca43124a686e9b1fd0e33483173375e5/src/components/ProductDetail/index.js#L8-L10

kuisathaverat: I workaround this random fail in the Docker image by adding this config to the RUM agent https://github.com/elastic/apm-integration-testing/blob/master/docker/opbeans/frontend_nginx/rum-config.template

sqren: Most of the opbeans apps add random failures to simulate real applications. 
I could be wrong but the random error seen here should only cause an uncaught exception.   It shouldn't cause Chrome to timeout like we see.

afaict the fix by @v1v works. You _could_ add a guard like:
```ts
} catch (e) {
  if (e.message.includes("timed out after")) {
    return run();
  }
  throw e;
}
```

docker/opbeans/rum/tasks.js
```diff
@@ -37,4 +37,11 @@ async function run() {
...
 
-run().catch(console.error.bind(console))
+try {
+  run();
+} catch (e) {
+  if (e.message.includes("timed out after")) {
+    return run();
+  }
+  console.error.bind(console);
+}
```
sqren: What you had before was better. This will only catch the first timeout. Additional timeouts will still make the process crash.

docker/opbeans/rum/tasks.js
```diff
@@ -37,4 +37,11 @@ async function run() {
...
 }
 
-run().catch(console.error.bind(console))
+try {
+  run();
+} catch (e) {
+  if (e.message.includes("timed out after")) {
+    return run();
+  }
+  console.error.bind(console);
```
sqren: This doesn't do anything. Probably what you meant is:
```js
console.error(e);
```

sqren:
The following approach will make the process resistant to timeout errors (and other errors as well) without using recursion:

```
async function run() {
  const chromeless = new Chromeless({
    lauchChrome: true
  });

  for (;;) {
    try {
      url = await chromeless.goto(url).evaluate(baseUrl => {
        // activateLoadGeneration is defined in opbeans-frontend
        if (typeof window.activateLoadGeneration === "function") {
          console.log("Activating route change load generation");
          window.activateLoadGeneration();
        }

        var links = document.querySelectorAll('a[href^="/"]');
        if (links && links.length) {
          var i = Math.floor(Math.random() * links.length);
          return links[i].href;
        } else {
          return baseUrl;
        }
      }, baseUrl);
    } catch (e) {
      // this will catch the error, log it and let the process continue
      console.error(`Error occurred while evaluating ${url}:`, e);
    }
    console.log(url);
    await sleep(8000 + Math.floor(Math.random() * 10000));
  }
}

run().catch(console.error.bind(console));
```

docker/opbeans/rum/tasks.js
v1v: There was a discussion earlier to avoid using the recursion as that particular implementation could potentially end as an infinite loop, refers to:
- https://github.com/elastic/apm-integration-testing/pull/664#discussion_r345300191
- https://github.com/elastic/apm-integration-testing/pull/664#discussion_r345293432

No strong opinion about what's the best strategy though, probably I misunderstood what you meant in your comment: https://github.com/elastic/apm-integration-testing/pull/664#discussion_r346056629

@kuisathaverat  and @mdelapenya what're your thoughts if we keep the first approach with the suggested improvement : https://github.com/elastic/apm-integration-testing/pull/664#discussion_r346056629 ?

kuisathaverat: I'll buy the Soren solution https://github.com/elastic/apm-integration-testing/pull/664#issuecomment-553860528 😃

kuisathaverat: BTW https://twitter.com/johnwilander/status/1176457013305303040?lang=en

PR #670, opened 2019-11-20
Configurable agent config polling interval.

### LINKED ISSUE #662: Make agent config frequency configurable
Should revert to old default of 30 secs, and keep the current 1 sec for automated tests.
Requested by @beniwohli

jalvz:
Adds an option to configure agent config polling interval.
This also reverts to the 30secs default and uses the option to override it on automatic tests.

Fixes #662


jalvz:
fyi @beniwohli

PR #673, opened 2019-11-22
feat: grab elasticsearch data on fails

kuisathaverat:
## What does this PR do?

Enable the feature to dump all data on Elasticsearch after a test failure. It also refactors some environment variables.

## Why is it important?

Some times the test fails because of Elasticsearch query returns less or more data than expected but we do not really know what it is in Elasticsearch, this PR grab that data so we can check what it is really on Elasticsearch.

### DIFF
--- a/.ci/integrationTestDownstream.groovy
+++ b/.ci/integrationTestDownstream.groovy
@@ -100,6 +100,7 @@ pipeline {
       environment {
         TMPDIR = "${WORKSPACE}"
         REUSE_CONTAINERS = "true"
+        PATH = "${WORKSPACE}/${BASE_DIR}/.ci/scripts:${env.PATH}"
       }
       steps {
         deleteDir()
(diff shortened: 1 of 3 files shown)


.ci/integrationTestDownstream.groovy
```diff
@@ -205,10 +209,7 @@ def runScript(Map params = [:]){
   unstash "source"
   dir("${BASE_DIR}"){
     withEnv(env){
-      sh """#!/bin/bash
-      export TMPDIR="${WORKSPACE}"
-      .ci/scripts/${agentType}.sh
-      """
+      sh(label: "Testing ${agentType}", script: ".ci/scripts/${agentType}.sh")
```
mdelapenya: Why adding the `.ci/scripts` to PATH if we use the relative path? Is there any other use case?

kuisathaverat: to call `elasticdump` from python we use the command directly, and it is OK you can use a local instance of elasticdump when you run the test locally

mdelapenya: Cool! I did not see the connection with `elasticdump` from this changeset.

Thanks!

v1v: ❤️

.ci/integrationTestSelector.groovy
```diff
@@ -123,6 +125,8 @@ pipeline {
       }
       environment {
         TMPDIR = "${WORKSPACE}"
+        ENABLE_ES_DUMP = "true"
+        PATH = "${WORKSPACE}/${BASE_DIR}/.ci/scripts:${env.PATH}"
```
v1v: Is it possible to move this section in the top-level environment?

kuisathaverat: `WORKSPACE` and `PATH` are Agent dependent environment variables it could change, even do 99.99999% of times would be the same there is a possibility to not.

v1v: Got it. Thanks

PR #675, opened 2019-11-27
documented the default users created by compose.py

beniwohli:
## What does this PR do?

Adds documentation for the default users

## Why is it important?

Because Slack isn't documentation ;p


graphaelli:
Thanks @beniwohli!

PR #678, opened 2019-11-28
fix: define ELASTIC_APM_VERIFY_SERVER_CERT only on Python latest versions

### LINKED ISSUE #685: Identify the root cause of the Python issues
since 2 weeks the Python tests on the master branch are failing, it does not happen locally so it is something related to the CI.

kuisathaverat:
## What does this PR do?

set Python 3.7 as the version used on containers, and define ELASTIC_APM_VERIFY_SERVER_CERT only on Python master branch. Finally, we have bump the APM agent Python versions to current latest minor releases

## Why is it important?

We have found two errors on current Python configuration for the integration test, Docker containers throw some errors with python 3.8 so we changed to the tested version 3.7, and ELASTIC_APM_VERIFY_SERVER_CERT is only supported on latest versions see https://github.com/elastic/apm-agent-python/issues/570

## Related issues
Closes #685

### DIFF
--- a/docker/python/django/Dockerfile
+++ b/docker/python/django/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3
+FROM python:3.7
 
 RUN pip install -q -U Django==2.1.5
 
--- a/docker/python/flask/Dockerfile
+++ b/docker/python/flask/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3
+FROM python:3.7
 
 RUN pip install -q -U Flask blinker gunicorn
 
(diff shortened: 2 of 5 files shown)


jalvz:
i approved a bit fast actually, python tests are still failing :thinking:

kuisathaverat:
@jalvz there is something else I am checking what, it is related also with Kibana, and I think that the watcher config but I am not sure yet.

v1v:
I'm not sure IIUC the context for this PR, is this a temporary fix or just to test a particular scenario within the PR itself?

kuisathaverat:
>I'm not sure IIUC the context for this PR, is this a temporary fix or just to test a particular scenario within the PR itself?

forget about is a PR to try to find the RC of the issue.

Issue #681, opened 2019-12-05
opbeans-dotnet does not have frontent 


gregkalapos:
On `apm.elstc.co`, which I think runs via `apm-integration-testing` `opbeans-dotnet` does not have a frontend, so all we see is very few requests to `GET /root/get`, nothing else.

Testing it by starting this repo locally, same - no frontend for `opbeans-dotnet`. 

The [`dockerfile`](https://github.com/elastic/apm-integration-testing/blob/master/docker/opbeans/dotnet/Dockerfile#L21) for `opbeans-dotnet` in this repo already has something for it, but the frontend does not seem to start up.

Discussed with @kuisathaverat  - probably we can either make the static serving of the frontend in the `opbeans-dotnet` itself work - I'm happy to help with this; or, another idea @kuisathaverat mentioned is to just host it in a side container.

PR #682, opened 2019-12-05
Make opbeans-dotnet listen on port 3000 instead of 80

gregkalapos:
Addressing #681 - which is that the load generator does not hit opbeans-dotnet except the health check url - so no real load was generated for opbeans-dotnet. - #681 talks about frontend, but turned out it is not really needed with the load generator. The current problem is that basically `opbeans-dotnet` does not generate any data, which this PR seems to solve.

@beniwohli found that opbeans-dotnet was listening on port `80` while the load generator uses port 3000 - or 3004 in case of .NET. 

In any case, this makes it work on my machine:
image


gregkalapos:
@v1v just pinging you fyi, because I think you worked on this part.

gregkalapos:
> Good catch! :S
> 
> I guess I was quite a newbie in one way or another that I didn't realize this was required

No worries. Just wanted to make sure I don't break anything :)

PR #686, opened 2019-12-09
enable apm-server self profiling by default

graphaelli:


graphaelli:
realized tests need to disable this, closing while making updates

graphaelli:
nm, they have `--no-apm-server-self-instrument` set already

scripts/modules/elastic_stack.py
```diff
@@ -52,8 +52,13 @@ def __init__(self, **options):
             ("monitoring.elasticsearch" if self.at_least_version("7.2") else "xpack.monitoring.elasticsearch", "true"),
             ("monitoring.enabled" if self.at_least_version("7.2") else "xpack.monitoring.enabled", "true")
         ])
-        if options.get("apm_server_self_instrument"):
+        if options.get("apm_server_self_instrument", True):
             self.apm_server_command_args.append(("apm-server.instrumentation.enabled", "true"))
+            if self.at_least_version("7.6") and options.get("apm_server_profile", True):
```
axw: are we backporting the feature to 7.6?

graphaelli: Since it's not a breaking a change I was expecting it to land in the next 7.x minor, so yes.  Any reservations about doing that?

graphaelli: Also, apm-server will still start up if we include the directive for now.  I included the version check to 1. to be explicit 2. support a future strict config mode

axw: :+1:  No reservations here

PR #691, opened 2019-12-12
replace Chromeless with puppeteer

beniwohli:
## What does this PR do?

Switches to Puppeteer for the Chrome-based load generation for RUM.

For now, we lose the remote debugging port at 9222, but I don't think anybody really uses that. Instead, I hooked up the page console to puppeteer's console, so those will appear in the docker logs, e.g.

    localtesting_7.5.0_opbeans-rum | PAGE LOG: log	Activating route change load generation	__puppeteer_evaluation_script__ (3:20)
    localtesting_7.5.0_opbeans-rum | PAGE LOG: warning	Synchronous XMLHttpRequest on the main thread is deprecated because of its detrimental effects to the end user's experience. For more help, check https://xhr.spec.whatwg.org/.	http://opbeans-python:3000/static/js/main.18f58ace.js (33:undefined)

## Why is it important?

Chromeless has been deprecated for a couple years, the cool kids use
Puppeteer these days.

As an added benefit, this should get rid of all the security alerts
that the old package-lock.json was triggering.

### DIFF
--- a/docker/opbeans/rum/Dockerfile
+++ b/docker/opbeans/rum/Dockerfile
@@ -1,38 +1,50 @@
-FROM node:8-slim
+FROM node:10-slim
 
-RUN apt-get -qq update && apt-get -qq install -y \
-	libgconf-2-4 \
-	apt-transport-https \
-	ca-certificates \
-	curl \
-	gnupg \
-	--no-install-recommends \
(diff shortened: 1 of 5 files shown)


vigneshshanmugam:
Thanks Beni 💃

PR #694, opened 2019-12-16
ci(jenkins): simplify the git base commit and avoid docker in the labels

v1v:
## What does this PR do?

- Remove docker label as it's redundant when using linux workers.
- Avoid the gitBaseCommit variable and use the env variable instead.

## Why is it important?

Let's keep the pipeline simpler.

## Related issues
Closes #ISSUE


v1v:
jenkins run the tests please

PR #695, opened 2019-12-16
Add flags combination to the docs

mdelapenya:
## What does this PR do?
This PR adds the most used flag combinations the team uses for internal development.

## Why is it important?
We want to share with the public this information just in case they need to replicate a custom environment we already use, not reinventing the wheel.

## Related issues
Closes elastic/observability-dev#131

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,8 @@ docker-compose stop <service name>
 
 ## Example environments
 
+We have a list with the most common flags combination that we internally use when developing our APM solution. You can find the list [here](https://docs.google.com/spreadsheets/d/1NCPR3RX670ZZOHXx7_y6bhe1zrSMYabPQ5Htheu3oZA/edit?usp=sharing).
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -79,6 +79,8 @@ docker-compose stop <service name>
 
 ## Example environments
 
+We have a list with the most common flags combination that we internally use when developing our APM solution. You can find the list [here](https://docs.google.com/spreadsheets/d/1NCPR3RX670ZZOHXx7_y6bhe1zrSMYabPQ5Htheu3oZA/edit?usp=sharing).
```
kuisathaverat: why a separate document? Could we move that info here to avoid maintaining a separate doc?

mdelapenya: Agreed. I'll move them here ;)

Thanks!

mdelapenya: I created the table, thanks for the feedback!

PR #703, opened 2019-12-26
Sort options and separate by newline

cachedout:
## What does this PR do?

This is a tiny DX improvement that makes the options list a bit easier to read.

## Why is it important?

When viewing options, this makes it easier to both grep for what I might want and it makes it a bit easier to scan through since options are now alphabetized.

## Related issues
No issue filed. Minor DX improvement.

### DIFF
--- a/scripts/modules/cli.py
+++ b/scripts/modules/cli.py
@@ -486,7 +486,7 @@ class LocalSetup(object):
         subprocess.call(cmd, shell=True)
 
     def listoptions_handler(self):
-        print("{}".format(" ".join(self.available_options)))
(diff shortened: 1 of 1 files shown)


cachedout:
@kuisathaverat I think I need to be added to a GH team in order to merge this. I think it's a protected branch.

kuisathaverat:
We would need to add you and @mikemadden42 to the team https://github.com/orgs/elastic/teams/observablt-robots/members 

cc @alvarolobato

cachedout:
Thanks, @kuisathaverat. I submitted a request to join the team. It looks like @alvarolobato is the maintainer so when he approves it, I should be set. :)

Issue #688, opened 2019-12-11
Ensure `service.version` changes from time to time


dgieselaar:
In https://github.com/elastic/kibana/pull/52640, we are displaying markers in the charts for when versions are "first seen", meaning the timestamp for the first event that has a distinct value for `service.version` per `service.name`. The opbeans apps are currently all reporting either no value for `service.version`, or the same one for all recorded events. This makes testing out of the box a little difficult because data needs to be augmented. 

Preferably the apm-integration-testing environment generates events with versions that change over time, so we see multiple versions for a given service.

beniwohli:
maybe we could set it to the agent version, now that most (or all?) opbeans variants get automatically built when a new version of the agent is released? Not sure how useful it would be though, as release frequency isn't that high.

kuisathaverat:
yep, this is the best option once we would have the Opbeans release process in place in every Opbean, as a workaround we can make that one of the Opbeans generate a random `service.version` on each build, this means that every time you launch the `compose.py` with the option `--force-build` this service name would change, Would it be enough for the moment @dgieselaar ?

dgieselaar:
I think it is. @sqren when apm.elstc.co is redeployed, do we pass the `--force-build` flag?

kuisathaverat:
checking how to implement it I have found that there is a flaw to change the service name for every opbeans, so it is implemented you only have to add a parameter like this one `--opbeans-go-service-name "opbeans-go-$RANDOM"` it would generate an environment variable like `ELASTIC_APM_SERVICE_NAME=opbeans-go-8391`

```
  --opbeans-dotnet-service-name OPBEANS_DOTNET_SERVICE_NAME
  --opbeans-dotnet01-service-name OPBEANS_DOTNET01_SERVICE_NAME
  --opbeans-go-service-name OPBEANS_GO_SERVICE_NAME
  --opbeans-go01-service-name OPBEANS_GO01_SERVICE_NAME
  --opbeans-java-service-name OPBEANS_JAVA_SERVICE_NAME
  --opbeans-java01-service-name OPBEANS_JAVA01_SERVICE_NAME
  --opbeans-python-service-name OPBEANS_PYTHON_SERVICE_NAME
  --opbeans-python01-service-name OPBEANS_PYTHON01_SERVICE_NAME
  --opbeans-ruby-service-name OPBEANS_RUBY_SERVICE_NAME
[...]

PR #709, opened 2020-01-03
Introdce APM_SERVICE_VERSION

cachedout:
## What does this PR do?

Introduces an option to pass the service version for APM agents.

## Why is it important?

This allows an arbitrary version to be passed in when running tests.

## Related issues
Refs https://github.com/elastic/apm-integration-testing/issues/688

### DIFF
--- a/scripts/tests/localsetup_tests.py
+++ b/scripts/tests/localsetup_tests.py
@@ -813,6 +813,17 @@ class LocalTest(unittest.TestCase):
         """)  # noqa: 501
         self.assertDictEqual(got, want)
 
+    def test_version_options(self):
+        docker_compose_yml = stringIO()
+        image_cache_dir = "/foo"
+        with mock.patch.dict(LocalSetup.SUPPORTED_VERSIONS, {'master': '8.0.0'}):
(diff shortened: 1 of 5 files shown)


kuisathaverat:
Could you add a test to check that it is possible to set the value by setting the parameter `--opbeans-service-version 1.2.3`?

cachedout:
@kuisathaverat Absolutely! Please take a look and see if the test approach that I used looks good to you.

scripts/tests/localsetup_tests.py
```diff
@@ -808,6 +813,17 @@ def test_start_6_3_default(self):
         """)  # noqa: 501
         self.assertDictEqual(got, want)
 
+    def test_version_options(self):
```
kuisathaverat: It is fine, there is a simple way to test the same by using `render` instead of the `LocalSetup`, something like this

```
def test_opbeans_java_service_version(self):
  opbeans = OpbeansJava(opbeans_java_service_version="1.2.3").render()["opbeans-java"]
  self.assertEqual("1.2.3", opbeans["environment"]["ELASTIC_APM_SERVICE_VERSION"])
```

cachedout:
jenkins re-test this

cachedout:
jenkins re-test this please

cachedout:
jenkins run the tests please

PR #710, opened 2020-01-07
ci(jenkins): env variables for pre-commit are not required

v1v:
## What does this PR do?

Remove unnecessary env variables

## Why is it important?

Simplify the pipeline 

## Related issues

Caused by https://github.com/elastic/apm-pipeline-library/pull/325

### DIFF
--- a/.ci/integrationTestSelector.groovy
+++ b/.ci/integrationTestSelector.groovy
@@ -2,7 +2,7 @@
 @Library('apm@current') _
 
 pipeline {
-  agent { label 'linux && immutable && docker' }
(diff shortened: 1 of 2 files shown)


.ci/integrationTestSelector.groovy
```diff
@@ -2,7 +2,7 @@
 @Library('apm@current') _
 
 pipeline {
-  agent { label 'linux && immutable && docker' }
+  agent { label 'linux && immutable' }
```
v1v: unrelated to this PR but let's help to avoid unnecessary labels.

PR #718, opened 2020-01-17
ci(jenkins): report downstream issues

v1v:
## What does this PR do?

Notify to the parentstream the reason for the downstream failures when they are related to a timeout or a test failure in the downstreams.

The notification is reported in the build description. Mitigate the timeout issues.

## Why is it important?

Easy to see timeout issues in the parent jobs or test failures.

## Related issues

Similar implementation to https://github.com/elastic/apm-agent-ruby/pull/682

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -131,6 +131,7 @@ def runJob(agentName, buildOpts = ''){
   def branch = env.CHANGE_ID?.trim() ? env.CHANGE_TARGET : env.BRANCH_NAME
   def job
   try {
+    def quiet = randomNumber(min: 2, max: 15) + 5
     job = build(job: "apm-integration-test-downstream/${branch}",
       parameters: [
       string(name: 'AGENT_INTEGRATION_TEST', value: agentName),
@@ -141,7 +142,7 @@ def runJob(agentName, buildOpts = ''){
       string(name: 'UPSTREAM_BUILD', value: currentBuild.fullDisplayName),
       booleanParam(name: 'DISABLE_BUILD_PARALLEL', value: '')],
       propagate: true,
-      quietPeriod: 10,
+      quietPeriod: quiet,
       wait: true)
   } catch(e) {
     job = e
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -1,6 +1,13 @@
 #!/usr/bin/env groovy
 @Library('apm@current') _
 
+import groovy.transform.Field
+
+/**
+ This is required to store the build status for the downstream jobs.
+*/
+@Field def itsDownstreamJobs = [:]
+
 pipeline {
   agent { label 'linux && immutable' }
   environment {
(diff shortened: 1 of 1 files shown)


.ci/Jenkinsfile
```diff
@@ -115,28 +122,31 @@ pipeline {
...
 }
 
 def runJob(agentName, buildOpts = ''){
-  def branch = env.BRANCH_NAME
-
-  if (env.CHANGE_ID) {
-    branch = env.CHANGE_TARGET
+  def branch = env.CHANGE_ID?.trim() ? env.CHANGE_TARGET : env.BRANCH_NAME
+  def job
+  try {
```
kuisathaverat: this block is repeatable on each job we check for build cause maybe we need a step to encapsulate it

```
catchDownstreamErrors('Downstream job name', registryVar) {
...
}
```

v1v: Not sure If I understand it correctly.

Something like the below snippet?

```
def runJob(agentName, buildOpts = ''){
   catchDownstreamErrors("Downstream job ${agentName}", itsDownstreamJobs) {
        def branch = env.CHANGE_ID?.trim() ? env.CHANGE_TARGET : env.BRANCH_NAME
        def quiet = randomNumber(min: 2, max: 15) + 5
        job = build(job: "apm-integration-test-downstream/${branch}",
        parameters: [
        string(name: 'AGENT_INTEGRATION_TEST', value: agentName),
        string(name: 'ELASTIC_STACK_VERSION', value: params.ELASTIC_STACK_VERSION),
        string(name: 'INTEGRATION_TESTING_VERSION', value: "${env.GIT_BASE_COMMIT}"),
        string(name: 'MERGE_TARGET', value: "${branch}"),
        string(name: 'BUILD_OPTS', value: "${params.BUILD_OPTS} ${buildOpts}"),
        string(name: 'UPSTREAM_BUILD', value: currentBuild.fullDisplayName),
        booleanParam(name: 'DISABLE_BUILD_PARALLEL', value: '')],
        propagate: true,
        quietPeriod: quiet,
        wait: true)
    }
}


// Step catchDownstreamErrors

call( ... ) {
   try {
       body()
    } catch(e) {

   } finally {
       // add the buildInfo details
   }
```

If so, I'm not sure if we can get the buildInfo details easily

kuisathaverat: yep that exactly, we can access to the data we can on the pipeline

.ci/Jenkinsfile
```diff
@@ -115,28 +122,32 @@ pipeline {
...
+    def quiet = randomNumber(min: 2, max: 15) + 5
```
cachedout: Is there any case where we might want to parametrize the span length? Perhaps in the future we may wish to set a certain repo to have a more or less aggressive retry policy?

v1v: This is a workaround to add some latency when running a bunch of downstream in parallel at the same time. It's not ideal though and IMO, it should be handled by the infra itself ecosystem, to avoid this weird and unnecessary code.

There were different options, in fact, https://github.com/elastic/apm-integration-testing/blob/e38ce906b1cdae87364bad03975fc31819013c17/.ci/Jenkinsfile#L22 is the programmatic way to avoid one single repo to cope with all the resources from the CI.

Since the retry/sleep workaround, when running the checkout, is now embedded with the shared library, I'd say I can remove this sleep randomness feature as it was an attempt to help with the timeout environmental issue, it did work just a bit by the way :(

cachedout: Makes sense. 👍 Thanks!

PR #722, opened 2020-01-21
Add Jaeger options to APM Server and expose default Jaeger ports.

jalvz:


kuisathaverat:
you are trying to escape without a test 😸, I think that something like this one it is enough https://github.com/elastic/apm-integration-testing/blob/master/scripts/tests/service_tests.py#L385-L390

jalvz:
haha, yes I know ;)

simitt:
Thanks for adding this @jalvz! I would even go further and enable it by default.

graphaelli:
+1 on @simitt proposal to enable by default.  I'd also support a single option that give both interfaces at once, perhaps `--no-apm-server-jaeger`

jalvz:
My intention was to add the Jaeger demo app pointing to apm-server, which probably we don't want by default. 
I will flip the logic and write a couple of doc lines for how to run the demo instead.

README.md
```diff
@@ -186,6 +186,15 @@ Omit `--skip-download` to just download images.
 
 `compose.py` includes unittests, `make test-compose` to run.
 
+### Jaeger
+
+APM Server can work as a drop-in replacement for a Jaeger collector, and ingest traces directly from a Jaeger client via
+HTTP/Thrift or from a Jaeger agent via gRPC.
+
+To test it with a Jaeger microservice demo, run separately:
+`docker run --rm -it -p8080-8083:8080-8083 -e JAEGER_ENDPOINT=http://apm-server:14268/api/traces  --network apm-integration-testing  jaegertracing/example-hotrod:1.16  all`
```
graphaelli: 👍 Great addition

Issue #726, opened 2020-01-23
Moving to Python 3


kuisathaverat:
Python 2.7 EOL arrived so we have to move to Python 3.

PR #733, opened 2020-01-23
feat: moving to Python 3

kuisathaverat:
## What does this PR do?

Move the Makefile to use Python3

## Why is it important?

Python 2.7 does not have support anymore.

## Related issues
Related to #726

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,6 +69,11 @@ test-agent-%: venv
 test-compose: venv
 	pytest scripts/tests/*_tests.py -v -s $(JUNIT_OPT)/compose-junit.xml
 
+test-compose-2:
+	virtualenv --python=python2.7 venv2
+	./venv2/bin/pip2 install mock pytest pyyaml
+	./venv2/bin/pytest --noconftest scripts/tests/*_tests.py
+
 test-kibana: venv
 	pytest tests/kibana/test_integration.py -v -s $(JUNIT_OPT)/kibana-junit.xml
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -31,7 +30,8 @@ all: test
 # The tests are written in Python. Make a virtualenv to handle the dependencies.
 # make doesn't play nicely with custom VENV, intended only for CI usage
 venv: requirements.txt
-	test -d $(VENV) || virtualenv -q --python=$(PYTHON3) $(VENV);\
+	$(PYTHON) -m venv $(VENV);\
```
v1v: Will it work if python3 is not available? The previous setup I guess it did allow that particular use case

kuisathaverat: no more support for Python2

Makefile
```diff
@@ -69,11 +69,6 @@ test-agent-%: venv
 test-compose: venv
 	pytest scripts/tests/*_tests.py -v -s $(JUNIT_OPT)/compose-junit.xml
 
-test-compose-2:
```
graphaelli: why remove this?  yes python2 is EOL but so many system still ship with it, if possible would be nice to keep compat for stop/start

Makefile
v1v: IIUC, the virtualenv installation used to be installed to support python3, that's not anymore the use case, as it does use the python3 binary to install the venv. In other words, I'll keep the previous command:

`test -d $(VENV) || virtualenv -q --python=$(PYTHON3) $(VENV);\`

The above command is backward compatible independently of the python version installed in the system, therefore, it's not necessary to set the PYTHON variable as it will work out of the box.  

I prefer to keep the functionality as it was even though the only required change could be de predefined python3 binary.

WDYT?

Makefile
kuisathaverat: we could keep it for a while

PR #735, opened 2020-01-23
feat: YAML output.

### LINKED ISSUE #712: Improve readability of generated docker-compose.yml
The docker-compose.yml file that is currently generated by compose.py is JSON, a subset of YAML. While it's technically YAML, it's not the best kind of YAML – it's hard to read, and hard to edit by hand.

For this, I propose we introduce an optional dependency on PyYAML, and add a flag to output in YAML. Ideally the output would look like a typical hand-crafted file, with the version at the top, volumes at the bottom, etc.

Another thing that makes the output difficult to read is the inclusion of redundant config, and config only relevant to running integration tests. It would be nice if we could generate a minimal docker-compose.yml for other purposes, such as in "quick start" documentation (https://github.com/elastic/apm-server/pull/3106).

For this I propose we stop explicitly specifying default config, and provide another new flag to set only the minimally required config for starting up the Stack and APM Server.

See https://gist.github.com/axw/05e6fb7f97108cb752b3051a0b304d2b as an example.

kuisathaverat:
## What does this PR do?

It changes the compose.py output to YAML.

## Why is it important?

The original JSON output is difficult to read and manipulate manually, the YAML output is the common format of docker-compose files and all the people are used to it.

BTW: I particularly hate YAML but in this case I think is better than the current output, maybe I am used to YAML 😢 

## Related issues
Closes elastic/apm-integration-testing#712

### DIFF
--- a/scripts/modules/cli.py
+++ b/scripts/modules/cli.py
@@ -9,6 +9,7 @@ import os
 import subprocess
 import sys
 import re
+import yaml
 
 from .beats import BeatMixin
 from .helpers import load_images
@@ -565,7 +566,10 @@ class LocalSetup(object):
             ),
         )
         docker_compose_path = args["docker_compose_path"]
-        json.dump(compose, docker_compose_path, indent=2, sort_keys=True)
+        yaml.dump(compose, docker_compose_path,
+            explicit_start = True,
+            default_flow_style=False,
+            indent = 2)
         docker_compose_path.flush()
 
         # try to figure out if writing to a real file, not amazing
--- a/scripts/modules/cli.py
+++ b/scripts/modules/cli.py
@@ -10,11 +10,6 @@ import subprocess
 import sys
 import re
 
-try:
-    import yaml
-except ImportError:
-    print("No able to load yaml lib, use '--output-format json'")
-
 from .beats import BeatMixin
 from .helpers import load_images
 from .opbeans import OpbeansService, OpbeansRum
@@ -387,7 +382,7 @@ class LocalSetup(object):
         parser.add_argument(
             '--output-format',
             help='Select the output format for the docker-compose.yml file. [yaml, json]',
-            default="yaml"
+            default="json"
         )
 
         self.store_options(parser)
@@ -578,12 +573,16 @@ class LocalSetup(object):
         docker_compose_path = args["docker_compose_path"]
 
         if args.get("output_format") == 'yaml':
+            try:
+                import yaml
+            except ImportError:
+                print("Failed to import 'yaml': pip install yaml, or specify an alternative --output-format.")
             yaml.dump(compose, docker_compose_path,
                       explicit_start=True,
                       default_flow_style=False,
                       indent=2
                       )
-        else:
+        elif args.get("output_format") == 'json':
             json.dump(compose, docker_compose_path, indent=2, sort_keys=True)
         docker_compose_path.flush()
 
--- a/scripts/modules/cli.py
+++ b/scripts/modules/cli.py
@@ -9,7 +9,11 @@ import os
 import subprocess
 import sys
 import re
-import yaml
+
+try:
+    import yaml
+except ImportError:
+    print("No able to load yaml lib, use '--output-format json'")
 
 from .beats import BeatMixin
 from .helpers import load_images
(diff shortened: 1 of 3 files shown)


graphaelli:
We've avoided this so far to remove any dependencies on 3rd party libs - json is in the stdlib, yaml is not.  How about an option to write the config as yaml, which only if invoked imports yaml, instead?

axw:
++ to what Gil said. e.g. add an "--output=yaml|json" flag or something like that.

Also I'd like to use OrderedDict to format the YAML idiomatically, but I can do that as a follow up.

mdelapenya:
👍 to an environmental configuration for the output

kuisathaverat:
pyYAML is in the prerequisites file, indeed tests do not run without it, so about to conditional load of the library, I'd said that it is a requirement thus cannot be conditional. 
About to have a flag to output the docker-compose in JSON we can implement it, but I guess nobody will use it

graphaelli:
> pyYAML is in the prerequisites file, indeed tests do not run without

Agreed, my concern is specifically about continuing to have `compose.py start` work without any libs outside of the stdlib

> a flag to output the docker-compose in JSON

I'm proposing the opposite, a flag to output in yaml so by default starting an environment works with just the standard library.

scripts/modules/cli.py
```diff
@@ -10,6 +10,11 @@
 import sys
 import re
 
+try:
+    import yaml
+except ImportError:
+    print("No able to load yaml lib, use '--output-format json'")
```
axw: This seems a bit too noisy for folks without PyYAML, if they're only interested in JSON output to begin with.

I suggest moving `import yaml` under `if args.get("output_format") == 'yaml':`, so there's only error output if YAML output format is specified. Error message should be something like "Failed to import 'yaml': pip install yaml, or specify an alternative --output-format.".

scripts/modules/cli.py
```diff
@@ -379,6 +384,12 @@ def init_start_parser(self, parser, services, argv=None):
             default="false"
         )
 
+        parser.add_argument(
+            '--output-format',
+            help='Select the output format for the docker-compose.yml file. [yaml, json]',
+            default="yaml"
```
axw: Can we defer changing the default format till later? Having the additional option is great, but changing behaviour might break peoples' usage.

scripts/modules/cli.py
```diff
@@ -565,7 +576,15 @@ def start_handler(self):
...
         docker_compose_path = args["docker_compose_path"]
-        json.dump(compose, docker_compose_path, indent=2, sort_keys=True)
+
+        if args.get("output_format") == 'yaml':
+            yaml.dump(compose, docker_compose_path,
+                      explicit_start=True,
+                      default_flow_style=False,
+                      indent=2
+                      )
+        else:
```
cachedout: Nit: This block will be entered if anything other than `yaml` is specified or no option is specified and the default is used. This may include misspellings such as `yamll`, for example, which should throw an error instead of silently switching to `json`, IMHO. Perhaps replace with an `elif`?

kuisathaverat:
I've changed it to use JSON by default, the `import yaml` is now in the YAML choose, finally y have changed the message.

IMHO the default output should be YAML and the `import yaml` should be mandatory,  less spaghetti code, easy to maintain, but I want to merge this and I understand all POV but breaking changes happens and should happen is the way to move forward.

scripts/modules/cli.py
```diff
@@ -379,6 +379,12 @@ def init_start_parser(self, parser, services, argv=None):
             default="false"
         )
 
+        parser.add_argument(
+            '--output-format',
+            help='Select the output format for the docker-compose.yml file. [yaml, json]',
```
graphaelli:
```suggestion
            choices=("json", "yaml"),
            help='Select the output format for the docker-compose.yml file. [yaml, json]',
```

graphaelli:
> IMHO the default output should be YAML and the `import yaml` should be mandatory

I understand but still disagree - the lower the barrier to using this tool the more users will be able to stand up solid development / demo environments.  I could see eventually making yaml the default output *if yaml is available* so something like:

```python
dumper = json.dumps
try:
  import yaml
  dumper = yaml.dumps
except ImportError:
  pass
[...]
dumper(compose, docker_compose_path, indent=2, sort_keys=True)
```

but am happy with the current changes as is.

PR #737, opened 2020-01-23
feat: test APM Server with TLS and a Token configured

### LINKED ISSUE #13: Add agent tests for different apm-server config.

kuisathaverat:
## What does this PR do?

It adds the flags needed to configure the APM Server to use TLS and a Token.

## Why is it important?

This will test the security configuration on the APM Server and Agents, the flags are added to the `ALL` test so it applies to all agents.

## Related issues
Closes elastic/apm-integration-testing#13 elastic/apm-integration-testing#221

### DIFF
--- a/.ci/scripts/all.sh
+++ b/.ci/scripts/all.sh
@@ -6,6 +6,23 @@ test -z "$srcdir" && srcdir=.
 # shellcheck disable=SC1090
 . "${srcdir}/common.sh"
 
-DEFAULT_COMPOSE_ARGS="${ELASTIC_STACK_VERSION} ${BUILD_OPTS} --no-apm-server-dashboards --no-apm-server-self-instrument --with-agent-rumjs --with-agent-dotnet --with-agent-go-net-http --with-agent-nodejs-express --with-agent-ruby-rails --with-agent-java-spring --with-agent-python-django --with-agent-python-flask --force-build --no-xpack-secure"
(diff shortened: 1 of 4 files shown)


.ci/scripts/all.sh
```diff
@@ -6,6 +6,21 @@ test -z "$srcdir" && srcdir=.
...
+  --no-apm-server-dashboards \
+  --no-apm-server-self-instrument \
+  --with-agent-rumjs \
+  --with-agent-dotnet \
+  --with-agent-go-net-http \
+  --with-agent-nodejs-express \
+  --with-agent-ruby-rails \
+  --with-agent-java-spring \
+  --with-agent-python-django \
+  --with-agent-python-flask \
```
mdelapenya: Small refactor for follow-ups: could we define an array of agents and loop them adding a `--with-agent-$agent`?

kuisathaverat: IMHO this would add an unnecessary complexity so add 8 values to a variable.

mdelapenya: SGTM 👍

kuisathaverat:
There are some test errors that we have to care about it

kuisathaverat:
for .NET we have to wait for https://github.com/elastic/apm-agent-dotnet/pull/465

kuisathaverat:
The only thing that still is not finished is the .Net agent support for ignoring certificates validation. https://github.com/elastic/apm-agent-dotnet/pull/465

PR #738, opened 2020-01-23
feat: flag to set the Kibana URL for the APM Server

### LINKED ISSUE #502: Add --apm-server-kibana-url flag
Similar to `--apm-server-elasticsearch-url`, it should update `apm-server.yaml`, specifically the `kibana` section, and also flip the `kibana.enabled` flag to `true`

Adapting this section is also needed for Agent Remote Configuration :tm:

kuisathaverat:
## What does this PR do?

It adds a new flag `--apm-server-kibana-url` to set the Kibana URL for the APM Server

## Why is it important?

Until now it was not possible you ave to change the docker-compose file manually.

## Related issues
Closes elastic/apm-integration-testing#502

### DIFF
--- a/scripts/modules/elastic_stack.py
+++ b/scripts/modules/elastic_stack.py
@@ -76,7 +76,7 @@ class ApmServer(StackService, Service):
             self.apm_server_command_args.extend([
                 ("apm-server.kibana.enabled", "true"),
                 ("apm-server.kibana.host",
-                    "{}".format(self.options.get("apm_server_kibana_url")))])
+                    self.options.get("apm_server_kibana_url", ""))])
             agent_config_poll = self.options.get("agent_config_poll", "30s")
             self.apm_server_command_args.append(("apm-server.agent.config.cache.expiration", agent_config_poll))
             if self.options.get("xpack_secure"):
--- a/scripts/modules/elastic_stack.py
+++ b/scripts/modules/elastic_stack.py
@@ -47,7 +47,7 @@ class ApmServer(StackService, Service):
             ("apm-server.write_timeout", "1m"),
             ("logging.json", "true"),
             ("logging.metrics.enabled", "false"),
-            ("setup.kibana.host", self.DEFAULT_KIBANA_HOST),
+            ("setup.kibana.host", "{}".format(self.options.get("apm_server_kibana_url"))),
             ("setup.template.settings.index.number_of_replicas", "0"),
             ("setup.template.settings.index.number_of_shards", "1"),
             ("setup.template.settings.index.refresh_interval", "1ms"),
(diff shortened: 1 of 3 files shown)


scripts/modules/elastic_stack.py
```diff
@@ -47,7 +47,7 @@ def __init__(self, **options):
             ("apm-server.write_timeout", "1m"),
             ("logging.json", "true"),
             ("logging.metrics.enabled", "false"),
-            ("setup.kibana.host", self.DEFAULT_KIBANA_HOST),
```
graphaelli: `setup.kibana.host` can actually be removed since we dropped 6.x support - that was used for loading dashboards into kibana.  You should set `apm-server.kibana.host` here instead:

https://github.com/elastic/apm-integration-testing/blob/33de17885a8bdc7fcc6b2d6f3b53ae318f403522/scripts/modules/elastic_stack.py#L78

mdelapenya: Also in the test? https://github.com/elastic/apm-integration-testing/pull/738/files#diff-13069991c5d90a2d74b80596f8686278R1407

kuisathaverat: done

scripts/tests/service_tests.py
```diff
@@ -697,8 +697,6 @@ def test_apm_server_acm(self):
         apm_server = ApmServer(version="7.3").render()["apm-server"]
         self.assertTrue("apm-server.kibana.enabled=true" in apm_server["command"],
                         "APM Server Kibana enabled by default")
-        self.assertTrue("apm-server.kibana.host=kibana:5601" in apm_server["command"],
```
kuisathaverat: it is tested in another test, also render does not grab default values do it fails.

scripts/modules/elastic_stack.py
```diff
@@ -76,7 +75,8 @@ def __init__(self, **options):
         elif self.at_least_version("7.3"):
             self.apm_server_command_args.extend([
                 ("apm-server.kibana.enabled", "true"),
-                ("apm-server.kibana.host", self.DEFAULT_KIBANA_HOST)])
+                ("apm-server.kibana.host",
+                    "{}".format(self.options.get("apm_server_kibana_url")))])
```
cachedout: Why is this `format()` necessary? Why not just:

`("apm-server.kibana.host", self.options.get("apm_server_kibana_url", ""))`

With the way it is written if the `get()` fails, the string will get set to the string `None`, which seems like it is probably not what we want?

kuisathaverat: I realized now why `("apm-server.kibana.host", self.options.get("apm_server_kibana_url"))` does no pass the tests, because the test that uses render does not apply the default values so the value assigned is None, so yes this is also valid and easy.

kuisathaverat:
```suggestion
                    self.options.get("apm_server_kibana_url", ""))])
```

PR #742, opened 2020-01-24
enable service maps ui by default

graphaelli:
works now that https://github.com/elastic/kibana/pull/55476 is in


scripts/modules/elastic_stack.py
```diff
@@ -702,6 +705,12 @@ def add_arguments(cls, parser):
             help="kibana elasticsearch output url(s)."
         )
 
+        parser.add_argument(
+            "--kibana-apm-disable-servicemaps",
```
jalvz: all the "don't do something" flags read like `--no-do-something`, what about following suit here?

jalvz: (asking because i always fail to remember `--skip-downlad` as a matter of fact :grimacing:

graphaelli: great call, updated

PR #759, opened 2020-02-14
opbeans-dotnet: support dotnet agent version >= 1.3 and < 1.3

v1v:
## What does this PR do?

If the version for the agent was not resolved as expected then it should fail:

```
+ '[' -z '' ']'
+ echo 'ERROR: DOTNET_AGENT_VERSION could not be calculated.'
+ exit 1
ERROR: DOTNET_AGENT_VERSION could not be calculated.
The command '/bin/sh -c ./run.sh' returned a non-zero code: 1
```

If the version for the agent was resolved but doesn't match then it should fail:

```
+ grep Elastic.Apm.NetCoreAll opbeans-dotnet.csproj
+ grep -i 'Version="11.3"'
ERROR: DOTNET_AGENT_VERSION mismatch
+ echo 'ERROR: DOTNET_AGENT_VERSION mismatch'
+ exit 1
The command '/bin/sh -c ./run.sh' returned a non-zero code: 1
```

## Why is it important?

Otherwise, the opbeans will not pick up the expected version but the one defined explicitly.

## Related issues

Closes https://github.com/elastic/apm-integration-testing/issues/758

PR #760, opened 2020-02-14
dotnet: support dotnet agent version >= 1.3 and < 1.3

v1v:
## What does this PR do?

If the version for the agent was not resolved as expected then it should fail:

```
+ echo 'ERROR: DOTNET_AGENT_VERSION could not be calculated.'
+ exit 1
ERROR: DOTNET_AGENT_VERSION could not be calculated.
The command '/bin/sh -c ./run.sh' returned a non-zero code: 1
```

If the version for the agent was resolved but doesn't match then it should fail:

```
## Force an error

+ grep -i 'Version="11.3"'
ERROR: DOTNET_AGENT_VERSION mismatch
+ echo 'ERROR: DOTNET_AGENT_VERSION mismatch'
+ exit 1
```

## Why is it important?

Otherwise, the dotnet app will not pick up the expected version but the one defined explicitly.

## Related issues

Leftover from https://github.com/elastic/apm-integration-testing/pull/759

### DIFF
--- a/docker/dotnet/run.sh
+++ b/docker/dotnet/run.sh
@@ -5,13 +5,14 @@ CSPROJ="TestAspNetCoreApp.csproj"
 
 PACKAGE=Elastic.Apm.NetCoreAll
 CSPROJ_VERSION="/src/dotnet-agent/src/Elastic.Apm.NetCoreAll/${PACKAGE}.csproj"
+BUILD_PROPS="/src/dotnet-agent/src/Directory.Build.props"
 
 if [ -z "${DOTNET_AGENT_VERSION}" ] ; then
 
-  git clone https://github.com/${DOTNET_AGENT_REPO}.git /src/dotnet-agent
+  git clone https://github.com/"${DOTNET_AGENT_REPO}".git /src/dotnet-agent
   cd /src/dotnet-agent
   git fetch -q origin '+refs/pull/*:refs/remotes/origin/pr/*'
-  git checkout ${DOTNET_AGENT_BRANCH}
+  git checkout "${DOTNET_AGENT_BRANCH}"
   git rev-parse HEAD
 
   ### Otherwise: /usr/share/dotnet/sdk/2.2.203/NuGet.targets(119,5): error : The local source '/src/local-packages' doesn't exist. [/src/dotnet-agent/ElasticApmAgent.sln]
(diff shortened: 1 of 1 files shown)


docker/dotnet/run.sh
```diff
@@ -5,13 +5,14 @@ CSPROJ="TestAspNetCoreApp.csproj"
 
 PACKAGE=Elastic.Apm.NetCoreAll
 CSPROJ_VERSION="/src/dotnet-agent/src/Elastic.Apm.NetCoreAll/${PACKAGE}.csproj"
+BUILD_PROPS="/src/dotnet-agent/src/Directory.Build.props"
 
 if [ -z "${DOTNET_AGENT_VERSION}" ] ; then
 
-  git clone https://github.com/${DOTNET_AGENT_REPO}.git /src/dotnet-agent
+  git clone https://github.com/"${DOTNET_AGENT_REPO}".git /src/dotnet-agent
```
v1v: lint fix

docker/dotnet/run.sh
```diff
@@ -5,13 +5,14 @@ CSPROJ="TestAspNetCoreApp.csproj"
...
   cd /src/dotnet-agent
   git fetch -q origin '+refs/pull/*:refs/remotes/origin/pr/*'
-  git checkout ${DOTNET_AGENT_BRANCH}
+  git checkout "${DOTNET_AGENT_BRANCH}"
```
v1v: lint fix

docker/dotnet/run.sh
```diff
@@ -21,10 +22,18 @@ if [ -z "${DOTNET_AGENT_VERSION}" ] ; then
 
   cd /src/aspnetcore
   mv /src/NuGet.Config .
+  # shellcheck disable=SC2016
```
v1v: lint fix

docker/dotnet/run.sh
```diff
@@ -21,10 +22,18 @@ if [ -z "${DOTNET_AGENT_VERSION}" ] ; then
...
+  DOTNET_AGENT_VERSION=$(grep 'PackageVersion' ${BUILD_PROPS} | sed 's#<.*>\(.*\)<.*>#\1#' | tr -d " ")
+  if [ -z "${DOTNET_AGENT_VERSION}" ] ; then
+    echo 'INFO: search version in the csproj. (only for agent version < 1.3)'
+    DOTNET_AGENT_VERSION=$(grep 'PackageVersion' ${CSPROJ_VERSION} | sed 's#<.*>\(.*\)<.*>#\1#' | tr -d " ")
+    if [ -z "${DOTNET_AGENT_VERSION}" ] ; then
+      echo 'ERROR: DOTNET_AGENT_VERSION could not be calculated.' && exit 1
+    fi
+  fi
+
+  dotnet add package ${PACKAGE} -v "${DOTNET_AGENT_VERSION}"
```
v1v: lint fix

PR #769, opened 2020-02-21
apm-ui: use the new kibana project layout

v1v:
## What does this PR do?

Fixes the current filesystem layout issue with the kibana project.
Reduce the verbose output in the clone
Fixes some linting issues.

## Why is it important?

Fix the apm-its

## Related issues

Leftovers from https://github.com/elastic/kibana/pull/57532
Closes https://github.com/elastic/apm-integration-testing/issues/770

## Questions

- [x] Do we need to support backward compatibility? 
- [ ] Do we need to backport this to the `*.x` branches?

### DIFF
--- a/scripts/kibana/validate-ts-interfaces-against-apm-server-sample-docs/scripts/clone-kibana.sh
+++ b/scripts/kibana/validate-ts-interfaces-against-apm-server-sample-docs/scripts/clone-kibana.sh
@@ -8,5 +8,14 @@ echo "Cloning Kibana: ${OWNER}:${BRANCH}"
 
 cd ./tmp || true
 git clone --quiet --depth 1 -b "${BRANCH}" "https://github.com/${OWNER}/kibana.git"
-mv ./kibana/x-pack/plugins/apm/typings/es_schemas ./apm-ui-interfaces
+
+### In 7.7 files moved around.
+### The below section keeps backward compatibility.
+oldLocation=./kibana/x-pack/legacy/plugins/apm/typings/es_schemas
+newLocation=./kibana/x-pack/plugins/apm/typings/es_schemas
+location=${oldLocation}
+if [ -d "${newLocation}" ] ; then
+   location=${newLocation}
+fi
+mv "${location}" ./apm-ui-interfaces
 rm -rf kibana
--- a/scripts/kibana/validate-ts-interfaces-against-apm-server-sample-docs/scripts/clone-kibana.sh
+++ b/scripts/kibana/validate-ts-interfaces-against-apm-server-sample-docs/scripts/clone-kibana.sh
@@ -4,6 +4,6 @@ BRANCH=${2:-master}
 echo "Cloning Kibana: $OWNER:$BRANCH"
 
 cd ./tmp
-git clone --depth 1 -b $BRANCH https://github.com/$OWNER/kibana.git
-mv ./kibana/x-pack/legacy/plugins/apm/typings/es_schemas ./apm-ui-interfaces
+git clone --quiet --depth 1 -b $BRANCH https://github.com/$OWNER/kibana.git
+mv ./kibana/x-pack/plugins/apm/typings/es_schemas ./apm-ui-interfaces
 rm -rf kibana
--- a/scripts/kibana/validate-ts-interfaces-against-apm-server-sample-docs/scripts/clone-kibana.sh
+++ b/scripts/kibana/validate-ts-interfaces-against-apm-server-sample-docs/scripts/clone-kibana.sh
@@ -6,7 +6,7 @@ BRANCH=${2:-master}
 
 echo "Cloning Kibana: ${OWNER}:${BRANCH}"
 
-cd ./tmp || true
+cd ./tmp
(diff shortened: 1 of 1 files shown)


scripts/kibana/validate-ts-interfaces-against-apm-server-sample-docs/scripts/clone-kibana.sh
```diff
@@ -1,9 +1,12 @@
+#!/usr/bin/env bash
```
v1v: Fix a lint defect

scripts/kibana/validate-ts-interfaces-against-apm-server-sample-docs/scripts/clone-kibana.sh
```diff
@@ -1,9 +1,12 @@
...
 OWNER=${1:-elastic}
 BRANCH=${2:-master}
 
-echo "Cloning Kibana: $OWNER:$BRANCH"
+echo "Cloning Kibana: ${OWNER}:${BRANCH}"
 
-cd ./tmp
-git clone --depth 1 -b $BRANCH https://github.com/$OWNER/kibana.git
-mv ./kibana/x-pack/legacy/plugins/apm/typings/es_schemas ./apm-ui-interfaces
+cd ./tmp || true
```
v1v: Fix a lint defect

scripts/kibana/validate-ts-interfaces-against-apm-server-sample-docs/scripts/clone-kibana.sh
```diff
@@ -1,9 +1,12 @@
...
+git clone --quiet --depth 1 -b "${BRANCH}" "https://github.com/${OWNER}/kibana.git"
```
v1v: - reduce verbose output in the CI.
- fix a lint defect

scripts/kibana/validate-ts-interfaces-against-apm-server-sample-docs/scripts/clone-kibana.sh
```diff
@@ -1,9 +1,12 @@
...
+mv ./kibana/x-pack/plugins/apm/typings/es_schemas ./apm-ui-interfaces
```
v1v: The real change!
Question:
- Do we need to support backward compatibility?

if so

```bash
newLocation=./kibana/x-pack/plugins/apm/typings/es_schemas
location=./kibana/x-pack/legacy/plugins/apm/typings/es_schemas
if [ -d "${newLocation}" ] ; then
   location=${newLocation}
fi
mv "${location} ./apm-ui-interfaces
```

scripts/kibana/validate-ts-interfaces-against-apm-server-sample-docs/scripts/clone-kibana.sh
sqren: What does this do? Is it if it fails navigating to `./tmp`?

scripts/kibana/validate-ts-interfaces-against-apm-server-sample-docs/scripts/clone-kibana.sh
sqren: Good question. If the above works I think that's a good approach. Perhaps add a comment saying that in 7.7 files moved around. Then we can move this again in a couple of minors from now.

scripts/kibana/validate-ts-interfaces-against-apm-server-sample-docs/scripts/clone-kibana.sh
v1v: Yes, have you got a different suggestion?

sqren: I'm wondering why this would fail - and fix the root cause of that.

scripts/kibana/validate-ts-interfaces-against-apm-server-sample-docs/scripts/clone-kibana.sh
sqren: I originally made this git clone a little more complicated than it has to be. What I wanted was a partial checkout of just a specific path (to make it faster). I didn't find any simple ways to do that. 
I don't think I saw any perf improvements by using the `--depth` flag (but not 100%)

scripts/kibana/validate-ts-interfaces-against-apm-server-sample-docs/scripts/clone-kibana.sh
v1v: I'll keep as it was then

v1v: rollback my changes

scripts/kibana/validate-ts-interfaces-against-apm-server-sample-docs/scripts/clone-kibana.sh
sqren: But you don't have to change this. Just thought I'd mention it.

Issue #770, opened 2020-02-21
[apm-ui] Tests are broken since a recent change in the Kibana apm-ui layout


v1v:
```
mv: cannot stat './kibana/x-pack/legacy/plugins/apm/typings/es_schemas': No such file or directory
Done in 25.13s.
yarn run v1.15.2
$ tsc ./tmp/apm-server-docs/*.ts --noEmit
tmp/apm-server-docs/ErrorRawDocs.ts(1,26): error TS2307: Cannot find module '../apm-ui-interfaces/raw/ErrorRaw'.
tmp/apm-server-docs/SpanRawDocs.ts(1,25): error TS2307: Cannot find module '../apm-ui-interfaces/raw/SpanRaw'
tmp/apm-server-docs/TransactionRawDocs.ts(1,32): error TS2307: Cannot find module '../apm-ui-interfaces/raw/TransactionRaw'.
error Command failed with exit code 1.
```

See build errors [here](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-integration-test-downstream/detail/master/4200/pipeline#step-81-log-18)

![image](https://user-images.githubusercontent.com/2871786/75034134-7014c300-54a4-11ea-83d9-c0a1b71230d3.png)

Caused by https://github.com/elastic/kibana/pull/57532

PR #771, opened 2020-02-21
[apm-ui] File names don't match the previous expected ones

v1v:
## What does this PR do?

Use the new files. It does not allow backward compatiblity 

## Why is it important?

Fix the APM-ITS for the UI

## Related issues

Leftover from https://github.com/elastic/apm-integration-testing/issues/770

### DIFF
--- a/scripts/kibana/validate-ts-interfaces-against-apm-server-sample-docs/scripts/download-sample-docs.ts
+++ b/scripts/kibana/validate-ts-interfaces-against-apm-server-sample-docs/scripts/download-sample-docs.ts
@@ -17,17 +17,17 @@ interface DocType {
 const docTypes: DocType[] = [
   {
     interfaceName: 'SpanRaw',
-    interfacePath: '../apm-ui-interfaces/raw/SpanRaw',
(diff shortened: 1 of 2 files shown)


scripts/kibana/validate-ts-interfaces-against-apm-server-sample-docs/scripts/download-sample-docs.ts
```diff
@@ -17,17 +17,17 @@ interface DocType {
 const docTypes: DocType[] = [
   {
     interfaceName: 'SpanRaw',
-    interfacePath: '../apm-ui-interfaces/raw/SpanRaw',
+    interfacePath: '../apm-ui-interfaces/raw/span_raw',
```
v1v: Master branch name file:
- https://github.com/elastic/kibana/tree/master/x-pack/plugins/apm/typings/es_schemas/raw

7.6 branch name file:
- https://github.com/elastic/kibana/tree/7.6/x-pack/legacy/plugins/apm/typings/es_schemas/raw

I've no clue if it's possible to support different interfaces at the same time.

Otherwise, the changes in master are breaking changes in this particular implementation. No clue how to move forward

Issue #782, opened 2020-02-26
Add support to enable the slow logs


kuisathaverat:
For some debug troubleshooting it is needed to have the slow logs of an index to find the problem. We need to add a way to enable the slow logs for an index, or all indices.

see https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules-slowlog.html

PR #788, opened 2020-03-03
Add Elasticsearch slowlog flag

cachedout:
## What does this PR do?

This adds the capability to pass in a flag called `--elasticsearch-slow-log` and have the slowlog automatically configured for Elasticsearch.

## Why is it important?

Improves ability to debug any issues.

## Related issues
Closes https://github.com/elastic/apm-integration-testing/issues/782

### DIFF
--- a/scripts/modules/helpers.py
+++ b/scripts/modules/helpers.py
@@ -82,6 +82,7 @@ def load_images(urls, cache_dir):
 DEFAULT_HEALTHCHECK_INTERVAL = "10s"
 DEFAULT_HEALTHCHECK_RETRIES = 12
 
+
 def _print_done(service_name):
     # 22 chars to ...
     done = '\033[32mdone'
@@ -92,6 +93,7 @@ def _print_done(service_name):
         s = s + ' '
     print("{service}{spaces}{dots} {done}".format(service=service_name, spaces=s, dots=dots, done=done))
 
+
 def try_to_set_slowlog():
     # This is a bit tricky to follow. What we're doing here is forking a "manager"
     # process that will occasionally attempt to configure the slow log. After it
@@ -99,6 +101,7 @@ def try_to_set_slowlog():
     # and terminates itself.
     manager_process = multiprocessing.Process(target=_set_slowlog_json)
     manager_process.start()
+
 
 def _set_slowlog_json():
     this_try = 0
@@ -147,6 +150,7 @@ def _set_slowlog_json():
             if json_ret.get("acknowledged"):
                 _print_done('Configuring slowlog')
                 break
+
 
 def curl_healthcheck(port, host="localhost", path="/healthcheck",
                      interval=DEFAULT_HEALTHCHECK_INTERVAL, retries=DEFAULT_HEALTHCHECK_RETRIES):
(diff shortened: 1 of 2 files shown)


scripts/modules/helpers.py
```diff
@@ -82,6 +83,75 @@ def load_images(urls, cache_dir):
...
+        this_try += 1
+        time.sleep(3)
+        completed_process = subprocess.run([
+            "curl",
+            "--fail-early",
+            "-f",
+            "--connect-timeout",
+            "1",
+            "-u",
+            "admin:changeme",
```
mdelapenya: Could this password overwritten by anybody else? Or is it always the same?

kuisathaverat: ouch!! I did not see it, yes it can be configurable

cachedout: Could somebody point me to how this might be done? It appears to me that the list of users [is pulled from here](https://github.com/elastic/apm-integration-testing/blob/master/docker/elasticsearch/users).

Searching for an option to overwrite this password doesn't yield anything obvious:

```
13:51 $ ./scripts/compose.py list-options|egrep pass
--apm-server-elasticsearch-password
--filebeat-elasticsearch-password
--heartbeat-elasticsearch-password
--metricbeat-elasticsearch-password
--packetbeat-elasticsearch-password
```

How does one overwrite the `admin` password on start, except to modify the `users` file in-lace?

kuisathaverat: `--apm-server-elasticsearch-password` will be enough, all the reset should be the same if the use the main Elasticsearch, if not, it does not matter.

cachedout: I'm still a little confused. Are you saying that if a user passes in the `--apm-server-elasticsearch-password` flag that Elasticsearch won't start with `admin:changeme` as specified in the `users` file?

kuisathaverat: if you pass `--apm-server-elasticsearch-password` you are probably not using the Elasticsearch instance in the `docker-compose` you ara using Elastic Cloud or ECK, so this admin user does not exists

scripts/modules/helpers.py
```diff
@@ -82,6 +83,74 @@ def load_images(urls, cache_dir):
...
+
+
+def _set_slowlog_json(password):
+    this_try = 0
+    tries = 30
+    while this_try <= tries:
+        this_try += 1
+        time.sleep(3)
+        completed_process = subprocess.run([
+            "curl",
```
sqren, line 113: Instead of forking a process could you do this in python - perhaps using the [Retry](https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#urllib3.util.retry.Retry) package?

Example:
```py
response = http.request(
  'PUT', 
  'localhost:9200/_settings?pretty', 
  retries=Retry(30)
  headers={'Content-Type': 'application/json'},
  body=json.dumps({...})
)
```

cachedout: IIRC, the reason we fell back to a shell invocation here is that we already have cURL installed and didn't want to introduce an additional Python dependency.

sqren: afaict `urllib` is already available: https://github.com/elastic/apm-integration-testing/blob/9221ec337e6e158f20cc0997a0804275d650d22d/requirements.txt#L38

But no biggie - I was just wondering when I saw this.

cachedout: Ah, you're right! Thanks for pointing this out. :)

PR #807, opened 2020-03-27
[WIP] feat: use kind instead GCP

kuisathaverat:
## What does this PR do?

Change the k8s provisioner to kind by default.

## Why is it important?

Kind provisions a k8s cluster on Docker, it does not have an additional cost for us. We can still use GCP by launching the job from Jenkins and selecting the `environment_file` parameter.


kuisathaverat:
see if https://kompose.io/ can help

PR #813, opened 2020-04-07
chore: use node:12 for RUM images

vigneshshanmugam:
+ Use Node.js 12 image instead of 8 as new packages are already deprecating support for 8 and the build fails with latest RUM master.


vigneshshanmugam:
Closing in favor of #815

PR #815, opened 2020-04-07
fix: RUM health check

kuisathaverat:
## What does this PR do?

* Bump the node version to 12 on the RUM containers.
* Change `apt-get` to `apt` command
* Fix the health check on the opbeans-rum Docker container.

## Why is it important?

If the port 9222 (chrome remote debugging) is not listening the health does not work.

### DIFF
--- a/docker/rum/Dockerfile
+++ b/docker/rum/Dockerfile
@@ -1,19 +1,19 @@
-FROM node:8-slim
+FROM node:12-slim
 
 ARG RUM_AGENT_BRANCH=master
 ARG RUM_AGENT_REPO=elastic/apm-agent-rum-js
 ARG APM_SERVER_URL
 
-RUN apt-get -qq update \
-    && apt-get -qq install -yq libgconf-2-4 \
-    && apt-get -qq install -y curl git --no-install-recommends \
-    && curl -L https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - \
+RUN apt -qq update \
+    && apt -qq install -yq libgconf-2-4 \
+    && apt -qq install -y curl git --no-install-recommends \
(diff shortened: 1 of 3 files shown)


docker/rum/Dockerfile
```diff
@@ -1,19 +1,17 @@
...
+    && apt install -qq -y curl git gnupg libgconf-2-4 --no-install-recommends \
+    && curl -sSfkL https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - \
     && sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list' \
-    && apt-get -qq update \
-    && apt-get -qq install -y google-chrome-unstable fonts-ipafont-gothic fonts-wqy-zenhei fonts-thai-tlwg fonts-kacst ttf-freefont python g++ build-essential \
+    && apt update -qq \
+    && apt install -qq -y google-chrome-unstable fonts-ipafont-gothic fonts-wqy-zenhei fonts-thai-tlwg fonts-kacst ttf-freefont python g++ build-essential \
       --no-install-recommends \
     && rm -rf /var/lib/apt/lists/* \
-    && apt-get -qq purge --auto-remove -y \
```
cachedout, line 16: Why remove the `purge` call?

kuisathaverat: we do not uninstall anything and the installations are made with `--no-install-recommends` so this command make nothing

```
       remove
           remove is identical to install except that packages are removed instead of installed.
           Note that removing a package leaves its configuration files on the system. If a plus
           sign is appended to the package name (with no intervening space), the identified
           package will be installed instead of removed.

       purge
           purge is identical to remove except that packages are removed and purged (any
           configuration files are deleted too).
```

PR #817, opened 2020-04-09
Opbeans java infer spans

graphaelli:
## What does this PR do?

Adds `--opbeans-java-infer-spans` option to enable [inferred span](https://www.elastic.co/guide/en/apm/agent/java/master/config-profiling.html#config-profiling-inferred-spans-enabled) collection for opbeans java.

image

## Why is it important?

This simplifies development and demo setup for inferred spans.

## Notes

To be flipped to on by default / switch option to `--no-` prefix when the visualization is improved.

### DIFF
--- a/scripts/modules/opbeans.py
+++ b/scripts/modules/opbeans.py
@@ -265,11 +265,18 @@ class OpbeansJava(OpbeansService):
             default=cls.DEFAULT_OPBEANS_VERSION,
             help=cls.name() + " version for the docker image of opbeans java"
         )
+        # only support by the java agent for now
+        parser.add_argument(
+            '--' + cls.name() + '-infer-spans',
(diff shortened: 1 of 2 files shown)


mdelapenya:
Hey @graphaelli LGTM, and I like to add if it i possible to set the configuration as a default instead, only creating the `--no-opbeans-java-infer-spans` flag if we want to skip it. This way we will maintain 50% of the code (only 1 flag). wdyt?

graphaelli:
@mdelapenya That's the intention in the longer term, per the description:

> To be flipped to on by default / switch option to `--no-` prefix when the visualization is improved.

mdelapenya:
Ah sorry, I understood the opposite, that we were going to have two flags 👍

Issue #781, opened 2020-02-25
Add support to run Agents using an ApiKey


kuisathaverat:
API key support is available from 7.6, so we have to add support to test the APM Server configured with an ApiKey that agents would use to authenticate.

* Start the Elastic Stack (Elasticsearch+Kibana+APM)
* Execute a couple of commands to grab an ApiKey
* Transform the ApiKey to base64
* Pass the ApiKey to all the agents with the environment variable ELASTIC_APM_API_KEY

This command prepares the user `apm` to generate an ApiKey

```
PUT /_security/privilege
{
 "apm": {
   "write_sourcemap": {
     "actions": [ "sourcemap:write" ]
   },
   "write_event": {
     "actions": [ "event:write" ]
   },
   "read_agent_config": {
     "actions": [ "config_agent:read" ]
   } }
}
```

This command requests an ApiKey

```
POST _security/api_key
{
   "name": "apm-backend",
   "role_descriptors": {
     "apm-backend": {
        "applications": [
           {
             "application": "apm",
             "privileges": ["*"],
             "resources": ["*"]
            }
          ]
        }
    }
}
```

This is the return JSON that contains the ApiKey

```
{
  "id" : "OEXB_m4B3SbfbZmPv8vb",
  "name" : "apm-backend",
  "api_key" : "XkZVRlCSSPWXM0IR0QdXDQ"
}
```
Finally we have to conver the ApiKey to base64 to use it
```
echo "XkZVRlCSSPWXM0IR0QdXDQ" |base64
```

related to https://github.com/elastic/apm/issues/183

kuisathaverat:
because we have to generate the ApiKey in runtime, probably the best option is to share a volume with the ApiKey between all the containers.

kuisathaverat:
once we have the parameter to run agents using ApiKey, we will need to run the `All` test using ApiKey, so we will have to add a new stage to test the ApiKey configuration.

v1v:
Some thoughts:
- Create a side container to prepare the context and notify the other agent containers.
- ~~Use the /etc/environment or /etc/profile.d (see [here](https://help.ubuntu.com/community/EnvironmentVariables#System-wide_environment_variables))~~ I'm afraid it's not POSIX and there are different number of containers that got different flavours.
- Use https://github.com/elastic/apm-integration-testing/pull/788 or something similar.
- Use `env-file` at runtime.

v1v:
### Manual testing with the env-file

[...]

PR #818, opened 2020-04-14
enable apikey for the apm-server and agents with a side container

v1v:
env_file is missing yet but it does not support runtime env variables

## Related issues

Relates to https://github.com/elastic/apm-integration-testing/issues/781


v1v:
This approach does not work. The reason is the environment variables cannot be changed during the docker-compose up.

PR #819, opened 2020-04-15
Enable APIKey in the agents

v1v:
## What does this PR do?

Enable the API key in the agents. For such it's required to:
- prepare the apm-its with running only the Elasticsearch service.
- run the api-key script
- then run the apm-its with all the required flags plus the new flag.

## Why is it important?

This is the approach we can enable it easily without much code.

## Related issues

Closes https://github.com/elastic/apm-integration-testing/issues/781

## Tasks
- [x] Fix linting
- [x] Add UTs

## Test

```
### Start the Elasticsearch instance
$ python scripts/compose.py start 8.0.0 --no-apm-server --no-kibana

### Wait for Elasticsearch to be available
$ sh -x scripts/create-api-key.sh

### Generate the API key
$ apiKey=$(scripts/create-api-key.sh)

### Run all the agents with EKA
$ python scripts/compose.py start 8.0.0  --elastic-apm-api-key ${apiKey} \
   --no-apm-server-dashboards --no-apm-server-self-instrument \
   --with-agent-rumjs \
   --with-agent-java-spring \
   --with-agent-ruby-rails \
   --with-agent-go-net-http \
   --with-agent-nodejs-express
```

### DIFF
--- a/scripts/create-api-key.sh
+++ b/scripts/create-api-key.sh
@@ -1,0 +1,38 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# shellcheck disable=SC2034
+privilege=$(curl -s -u admin:changeme -X PUT "localhost:9200/_security/privilege" -H 'Content-Type: application/json' -d'
+{
+"apm": {
+    "write_sourcemap": {
+      "actions": [ "sourcemap:write" ]
+    },
+    "write_event": {
+      "actions": [ "event:write" ]
+    },
+    "read_agent_config": {
+      "actions": [ "config_agent:read" ]
+    }
+  }
+}
+')
+
+apiKey=$(curl -s -u admin:changeme "localhost:9200/_security/api_key" -H 'Content-Type: application/json' -d'
+{
+  "name": "apm-backend",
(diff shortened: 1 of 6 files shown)


v1v:
This approach does not work. The reason is the environment variables cannot be changed during the docker-compose up.

scripts/modules/apm_agents.py
```diff
@@ -48,11 +57,7 @@ def _content(self):
             image=None,
             labels=None,
             logging=None,
-            environment={
```
v1v, line 51: This is required to be able to update the environments correctly. I did try with different approaches but there were certain issues with the environment type and how it's internally manipulated for other services.

scripts/create-api-key.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
```
mdelapenya, line 1: Will we need Windows support for this script?

v1v: Potentially, but if so, then we might come up with some follow-ups

PR #822, opened 2020-04-16
[7.x] Enable APIKey in the agents (#819)

v1v:
 - Enable APIKey in the agents (#819)


v1v:
This is not a trivial backport since there was a refactor back in the days and was not backported

PR #823, opened 2020-04-17
docker/apm-server: don't copy fields.yml

axw:
## What does this PR do?

fields.yml file is not required, as it gets embedded in the binary. This PR stops copying fields.yml into the Docker image used for integration testing.

## Why is it important?

I'm moving where fields.yml gets built in apm-server, which causes this to break -- may as well
just stop copying it in since it's not needed.

## Related issues

None.


axw:
Merging without review given the triviality of it. If there's a problem we can revert.

Issue #829, opened 2020-04-21
Add default parameter when running Kibana to opt-out from Telemetry


afharo:
We can see many telemetry reports in our Telemetry Production cluster from instances that are created and stopped on the same day and with the name `cluster_name: "docker-cluster"`. 

Can we add the option `--telemetry.optIn=false` in the Kibana service's start command to opt-out from sending any telemetry reports?

kuisathaverat:
Let's disable telemetry by default, and add a flag to enable it when we need it.

kuisathaverat:
Let's disable telemetry by default, and add a flag to enable it when we need it.

kuisathaverat:
I've checked it, we set the env var `XPACK_XPACK_MAIN_TELEMETRY_ENABLED`, but it has changed on https://github.com/elastic/kibana/commit/c308048f04e245080fa51e1be2cdc12668c33740 so I've tried to set a similar env var for the new setting, I didn't have too much luck, none of the following env vars disable the telemetry 😞 @afharo Do you know if there is an env var to disable telemetry?

```
        "xpack.telemetry.enabled": "false",
        "xpack_telemetry_enabled": "false",
        "xpackTelemetryEnabled": "false",
        "telemetryEnabled": "false",
        "telemetry.enabled": "false",
        "telemetry_enabled": "false",
        "XPACK_TELEMETRY_ENABLED": "false",
```

afharo:
I can see that env var comes from [here](https://github.com/elastic/kibana-docker/blob/master/.tedi/template/kibana-docker#L147). But that repo is archived and used only for releases pre-6.6 (according to its readme).

I tried to find the new version of it in the Kibana repo: https://github.com/elastic/kibana/blob/master/src/dev/build/tasks/os_packages/docker_generator/resources/bin/kibana-docker#L237

I guess you are looking for either `TELEMETRY_ENABLED` or `TELEMETRY_OPTIN` 😉

afharo:
Note that this only works for docker builds (which I believe it's your case anyway) 😃

PR #831, opened 2020-04-22
fix: disable telemetry on 7.5+

kuisathaverat:
## What does this PR do?

It adds an env var to disable telemetry on 7.6.

## Why is it important?

We send telemetry data from our tests and local development environments, this data is nois.

## Related issues
Closes https://github.com/elastic/apm-integration-testing/issues/829


scripts/modules/elastic_stack.py
```diff
@@ -725,6 +725,8 @@ def __init__(self, **options):
             self.environment["XPACK_MONITORING_ENABLED"] = "true"
             if self.at_least_version("6.3"):
                 self.environment["XPACK_XPACK_MAIN_TELEMETRY_ENABLED"] = "false"
+            if self.at_least_version("7.6"):
```
afharo, line 728: I'd say it's `7.5`.
This is the version where we added the _opted-in_ by default and I've checked in the `kibana-docker` file and this property exists for that version

PR #833, opened 2020-04-28
Build apm-server-oss when passing --oss

axw:
## What does this PR do?

When passing `--oss` to compose.py, make sure we build the `apm-server-oss` binary for the apm-server Docker image.

## Why is it important?

It's expected to run the OSS-only apm-server, and work when running with a pure-OSS stack.

## Related issues

None.

### DIFF
--- a/docker/apm-server/Dockerfile
+++ b/docker/apm-server/Dockerfile
@@ -1,7 +1,13 @@
 ARG apm_server_base_image=docker.elastic.co/apm/apm-server:8.0.0-SNAPSHOT
 ARG go_version=1.13
+ARG apm_server_binary=apm-server
+
+###############################################################################
+# Build stage: build apm-server binary and update apm-server.yml
+###############################################################################
 
 FROM golang:${go_version} AS build
+ARG apm_server_binary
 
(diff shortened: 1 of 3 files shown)


docker/apm-server/Dockerfile
```diff
@@ -1,7 +1,13 @@
 ARG apm_server_base_image=docker.elastic.co/apm/apm-server:8.0.0-SNAPSHOT
 ARG go_version=1.13
+ARG apm_server_binary=apm-server
+
+###############################################################################
+# Build stage: build apm-server binary and update apm-server.yml
+###############################################################################
 
 FROM golang:${go_version} AS build
+ARG apm_server_binary
```
kuisathaverat, line 10: by default, we can build the x-pack version

```suggestion
ARG apm_server_binary="apm-server"
```

axw: The default value comes from the global ARG at the top of the Dockerfile - it's not necessary to repeat it.

axw: (It is necessary to have an `ARG apm_server_binary` in each build stage though: https://github.com/moby/moby/issues/37345)

PR #820, opened 2020-04-15
Enable TLS for Kibana and Elasticsearch

v1v:
## What does this PR do?

Enable TLS for Kibana and Elasticsearch.
Refactor the curl healthcheck to support https protocol and `-k` by default.

## Why is it important?

Support TLS for Kibana and ES

## Related issues

Closes https://github.com/elastic/apm-integration-testing/issues/810

## Further details

The apm-server certificate cannot be reused since there is a certs validation 

```
{"type":"log","@timestamp":"2020-04-16T09:31:48+00:00","tags":["error","elasticsearch","admin"],"pid":7,"message":"Request error, retrying\nGET https://elasticsearch:9200/_nodes?filter_path=nodes.*.version%2Cnodes.*.http.publish_address%2Cnodes.*.ip => Hostname/IP does not match certificate's altnames: Host: elasticsearch. is not cert's CN: apm-server"}
```

### Start ES only with TLS

```
## Start with TLS
$ python scripts/compose.py start 8.0.0 --no-apm-server --no-kibana --elasticsearch-enable-tls

## Curl https -> 200
$ curl -k https://localhost:9200/ -I
HTTP/1.1 200 OK
content-type: application/json; charset=UTF-8
content-length: 543

## Curl http -> unreachable
$ curl http://localhost:9200/ -I
curl: (52) Empty reply from server

### Validate certificate
$ cat /etc/hosts | grep elasticsearch
127.0.0.1 elasticsearch

$ curl --cacert ./scripts/tls/ca/ca.crt  -u admin:changeme 'https://elasticsearch:9200/_cat/nodes?v'
ip         heap.percent ram.percent cpu load_1m load_5m load_15m node.role master name
172.28.0.2           16          60   1    0.28    0.26     0.24 dilmrt    *      8d00f8bebdaa
```

### Start Kibana and ES only with TLS and no certs validation

```
$ python scripts/compose.py start 8.0.0 --no-apm-server --with-kibana --elasticsearch-enable-tls --kibana-enable-tls --no-verify-server-cert 

```

### Start Kibana and ES only with TLS

```
### Start the Elasticsearch and Kibana instances with TLS
$ python scripts/compose.py start 8.0.0 --no-apm-server --elasticsearch-enable-tls --kibana-enable-tls

$ docker logs localtesting_8.0.0_kibana
[...]

PR #838, opened 2020-05-07
configure apm-server elasticsearch and kibana conns when TLS is enabled

graphaelli:
## What does this PR do?

Configures apm-server connections when TLS is enabled for elasticsearch and/or kibana.

## Why is it important?
APM Server is unable to communicate with Elaticsearch and Kibana when `--elasticsearch-enable-tls --kibana-enable-tls` are set, reporting:

```
Failed to connect to backoff(elasticsearch(http://elasticsearch:9200)): Get http://elasticsearch:9200: EOF

failed to obtain connection to Kibana: fail to get the Kibana version: HTTP GET request to http://kibana:5601/api/status fails: fail to execute the HTTP GET request: Get http://kibana:5601/api/status: EOF. Response: ."
```

The options are more commonly enabled now that alerting requires TLS configuration.

## Related issues
Follow up on #820

### DIFF
--- a/scripts/modules/elastic_stack.py
+++ b/scripts/modules/elastic_stack.py
@@ -20,6 +20,7 @@ class ApmServer(StackService, Service):
     DEFAULT_OUTPUT = "elasticsearch"
     OUTPUTS = {"elasticsearch", "file", "kafka", "logstash"}
     DEFAULT_KIBANA_HOST = "kibana:5601"
+    STACK_CA_PATH = "/usr/share/apm-serer/config/certs/stack-ca.crt"
 
     def __init__(self, **options):
         super(ApmServer, self).__init__(**options)
@@ -65,6 +66,9 @@ class ApmServer(StackService, Service):
             "enable_elasticsearch", True) else {}
         self.build = self.options.get("apm_server_build")
 
+        self.es_tls = self.options.get("elasticsearch_enable_tls", False)
+        self.kibana_tls = self.options.get("kibana_enable_tls", False)
+
         if self.options.get("apm-server-experimental-mode", True) and self.at_least_version("7.2"):
             self.apm_server_command_args.append(("apm-server.mode", "experimental"))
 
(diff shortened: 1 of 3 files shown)


scripts/modules/elastic_stack.py
```diff
@@ -153,8 +162,10 @@ def add_es_config(args, prefix="output"):
                     args.append((prefix + ".elasticsearch.{}".format(cfg), self.options[es_opt]))
                 elif self.options.get("xpack_secure"):
                     args.append((prefix + ".elasticsearch.{}".format(cfg), default_apm_server_creds.get(cfg)))
+            if tls:
+                args.append((prefix + ".elasticsearch.ssl.certificate_authorities", "['" + self.STACK_CA_PATH + "']"))
 
-        es_urls = self.options.get("apm_server_elasticsearch_urls") or [self.DEFAULT_ELASTICSEARCH_HOSTS]
```
mdelapenya, line 157: question: do we need to update the `self. DEFAULT_ELASTICSEARCH_HOSTS` usage across different services (like aux_services, opbeans and beats? They still use the variable

kuisathaverat: I gues so.

scripts/modules/elastic_stack.py
```diff
@@ -20,6 +20,7 @@ class ApmServer(StackService, Service):
     DEFAULT_OUTPUT = "elasticsearch"
     OUTPUTS = {"elasticsearch", "file", "kafka", "logstash"}
     DEFAULT_KIBANA_HOST = "kibana:5601"
+    STACK_CA_PATH = "/usr/share/apm-serer/config/certs/stack-ca.crt"
```
simitt, line 23: `%s/apm-serer/apm-server`

graphaelli: I prefer it this way

scripts/modules/elastic_stack.py
graphaelli: yes certainly.  We also need to apply these tls client settings to all beats, I'd prefer that be done in a follow up

kuisathaverat: if you do not have time open and issue and we take care of it :D

PR #843, opened 2020-05-25
Fix broken upload-sourcemap command

### LINKED ISSUE #830: upload-sourcemap command doesn't work
When running `./scripts/compose.py upload-sourcemap`, I receive the following error:
```
No source map found in apm-integration-testing/scripts/docker/opbeans/node/sourcemaps/*.map Start the opbeans-node container, it will create one automatically.
```

I have opbeans-node running, and after doing a search, it looks like opbeans-node is actually storing the generated source map here (no `/scripts/` in the path):
```
apm-integration-testing/docker/opbeans/node/sourcemaps/*.map
```

I copied the generated source map file to the location `upload-sourcemap` expects, and ran the command again:
```
Uploading apm-integration-testing/scripts/docker/opbeans/node/sourcemaps/main.adaa4c5b.js.map for client version 1.0.0
```

As you can see, there is no error this time, but it doesn't look like the upload works. The source map index is still nonexistent.

~I was able to upload the source map manually with curl, and while a document is created in the sourcemap index, it doesn't seem to contain the actual sourcemap.~

EDIT: I was able to manually upload the sourcemap and get it to work with this:
```
curl http://localhost:8200/assets/v1/sourcemaps -X POST \
  -F sourcemap="@./docker/opbeans/node/sourcemaps/main.adaa4c5b.js.map" \
  -F service_version="1.0.0" \
  -F service_name="client" \
  -F bundle_filepath="http://localhost:3000/static/js/main.adaa4c5b.js"
```

cachedout:
## What does this PR do?
Fixes broken `upload-sourcemap` command.

## Why is it important?

Previously, the `upload-sourcemap` command would not work because it both looked in the incorrect path for a sourcemap file and then failed to use the resolved path in the upload. This corrects both issues.

## Related issues
Closes #830

### DIFF
--- a/scripts/modules/cli.py
+++ b/scripts/modules/cli.py
@@ -660,7 +660,7 @@ class LocalSetup(object):
                 g = os.path.abspath(os.path.join(os.path.dirname(
                     __file__),
                     '../../docker/opbeans/node/sourcemaps/*.map')
-                    )
+                )
                 sourcemap_file = glob.glob(g)[0]
             except IndexError:
                 print(
--- a/scripts/modules/cli.py
+++ b/scripts/modules/cli.py
@@ -657,7 +657,7 @@ class LocalSetup(object):
                 sys.exit(1)
         else:
             try:
-                g = os.path.abspath(os.path.join(os.path.dirname(__file__), '../docker/opbeans/node/sourcemaps/*.map'))
+                g = os.path.abspath(os.path.join(os.path.dirname(__file__), '../../docker/opbeans/node/sourcemaps/*.map'))
                 sourcemap_file = glob.glob(g)[0]
             except IndexError:
                 print(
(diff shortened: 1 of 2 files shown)


scripts/modules/cli.py
```diff
@@ -689,7 +689,7 @@ def upload_sourcemaps_handler(self):
             '-F service_name="{service_name}" '
             '-F service_version="{service_version}" '
             '-F bundle_filepath="{bundle_path}" '
-            '-F sourcemap=@/tmp/sourcemap '
+            '-F sourcemap=@{sourcemap_file} '
```
v1v, line 692: What's the different between `@{filename}`  and  `@filename` syntax?

cachedout: This is a Python-ism. The curly-braces are indicating interpolation from the `format()` call that is chained to the end of the string. So, the value found in L699 and substituting it for `{sourcemap_file}`.

v1v: Forget about what I asked, for some reason I though it was bash-ish :( Thanks for the clarification :)

PR #849, opened 2020-06-04
Fetch .NET Agent version from VersionPrefix

gregkalapos:
Addressing https://github.com/elastic/apm-agent-dotnet/pull/869#issuecomment-638778829.

https://github.com/elastic/apm-agent-dotnet/pull/869 changes where the agent version is stored and this PR adapts the script that fetches the version.

I think this is a 🐔 🥚 problem - either we merge the PR in the .NET repo first and the IT that is triggered from the .NET repo will fail (as seen on the referenced comment) or we merge this and in case this runs before we merge .NET PR (not sure on merging this PR we run the whole .NET pipeline) then it'd fail again. To me merging this first sounds better.


v1v:
I'll provide some backward compatibility in case we would like to support an old version of the apm-agent-dotnet with the latest opbeans.

I'll raise a PR to your PR to support this :)

v1v:
here https://github.com/gregkalapos/apm-integration-testing/pull/1

gregkalapos:
> here [gregkalapos#1](https://github.com/gregkalapos/apm-integration-testing/pull/1)

Thanks, PR merged.

PR #854, opened 2020-06-08
[7.x] ci(docs): generate debug docs with the data (#814)

v1v:
 - ci(docs): generate debug docs with the data (#814)

### DIFF
--- a/.ci/integrationTestSelector.groovy
+++ b/.ci/integrationTestSelector.groovy
@@ -95,7 +95,9 @@ pipeline {
       }
       environment {
         TMPDIR = "${WORKSPACE}"
-        REUSE_CONTAINERS = "true"
+        ENABLE_ES_DUMP = "true"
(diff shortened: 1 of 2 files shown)


.ci/integrationTestSelector.groovy
```diff
@@ -95,7 +95,9 @@ pipeline {
       }
       environment {
         TMPDIR = "${WORKSPACE}"
-        REUSE_CONTAINERS = "true"
+        ENABLE_ES_DUMP = "true"
+        PATH = "${WORKSPACE}/${BASE_DIR}/.ci/scripts:${env.PATH}"
+        NAME = 'all'
```
mdelapenya, line 100: Is $NAME descriptive enough? Wdyt about $MODULE_NAME?

I know it's a backport and the forward version is in master, so if considered I'd apply it to master too

v1v: Uhm, it's not really a module though, but something else, `name` sounds generic enough for the current context

PR #856, opened 2020-06-08
[7.x] Fetch .NET Agent version from VersionPrefix (#849)

v1v:
 - Fetch .NET Agent version from VersionPrefix (#849)

### Issues
Depends on https://github.com/elastic/apm-integration-testing/pull/857


v1v:
I'll merge this PR as soon as possible to fix the current broken builds

PR #894, opened 2020-07-31
Introduce APM log level flag

cachedout:
## What does this PR do?

Introduces a new flag called `--apm-log-level` which is set to `info` out-of-the-box when running `compose.py` directly, but when invoked via the `.ci/scripts/*.sh` jobs for the APM agents, it is set to `debug` by default.

## Why is it important?

This was requested by the APM agent teams as a means to debug persistent failures in Elastic Cloud.

cc: @felixbarny

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -31,7 +31,7 @@ pipeline {
   }
   parameters {
     string(name: 'ELASTIC_STACK_VERSION', defaultValue: "8.0.0", description: "Elastic Stack Git branch/tag to use")
-    string(name: 'BUILD_OPTS', defaultValue: "", description: "Addicional build options to passing compose.py")
+    string(name: 'BUILD_OPTS', defaultValue: "", description: "Additional build options to pass to compose.py")
     booleanParam(name: 'Run_As_Master_Branch', defaultValue: false, description: 'Allow to run any steps on a PR, some steps normally only run on master branch.')
   }
   stages{
--- a/.ci/scripts/agent.sh
+++ b/.ci/scripts/agent.sh
@@ -14,6 +14,6 @@ DEFAULT_COMPOSE_ARGS="${ELASTIC_STACK_VERSION} ${BUILD_OPTS} \
   --no-apm-server-self-instrument \
   --apm-server-agent-config-poll=1s \
   --force-build --no-xpack-secure \
-  --apm-log-level debug"
+  --apm-log-level=debug"
 export COMPOSE_ARGS=${COMPOSE_ARGS:-${DEFAULT_COMPOSE_ARGS}}
 runTests "env-agent-${AGENT}" "docker-test-agent-${AGENT}"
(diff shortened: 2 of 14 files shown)


cachedout:
I started a few of these containers locally and verified that the environment seems to be correct.

For example, given this command `python3 scripts/compose.py start 7.0.0  --build-parallel   --no-apm-server-dashboards   --no-apm-server-self-instrument   --no-kibana   --with-agent-ruby   --force-build   --no-xpack-secure   --apm-log-level=debug` we see a container which looks like this upon entering:

```
❯ docker exec -it 3dbc47134eeb /bin/bash
root@3dbc47134eeb:/app/testapp# export
declare -x APM_SERVER_URL="http://apm-server:8200"
declare -x BUNDLE_APP_CONFIG="/usr/local/bundle"
declare -x BUNDLE_SILENCE_ROOT_WARNING="1"
declare -x ELASTIC_APM_API_REQUEST_TIME="3s"
declare -x ELASTIC_APM_LOG_LEVEL="debug"
declare -x ELASTIC_APM_SERVER_URL="http://apm-server:8200"
declare -x ELASTIC_APM_SERVICE_NAME="railsapp"
declare -x ELASTIC_APM_VERIFY_SERVER_CERT="false"
<snip>
```

This corresponds to what I believe the environment should look like given documentation such as this: https://www.elastic.co/guide/en/apm/agent/go/current/configuration.html#config-log-level

However, the documentation does not make it clear if whether or not _every_ agent respects this environment variable. 

@felixbarny Can you confirm that setting this environment variable should enforce debugging logging for all agents?

felixbarny:
The agents are not aligned on the possible values of the log level. However, most support `debug`. Ruby is an exception here. .NET supports `Debug` but I'm not sure if it's case sensitive. Python doesn't have the setting at all.

@mikker What are the possible values `ELASTIC_APM_LOG_LEVEL` for the Ruby agent? I couldn't quite get this info out of https://www.elastic.co/guide/en/apm/agent/ruby/1.x/configuration.html#config-log-level

mikker:
> What are the possible values ELASTIC_APM_LOG_LEVEL for the Ruby agent? 

They are the integers `0` to `5` (or strings converted to ints) as per https://www.rubydoc.info/stdlib/logger/Logger/Severity. We could definitely add a string => constant conversion for cross-agent alignment's sake.

beniwohli:
@cachedout for Python/Django, you could set the level here: https://github.com/elastic/apm-integration-testing/blob/master/docker/python/django/testapp/testapp/settings.py#L119

something like `os.environ.get("ELASTIC_APM_LOG_LEVEL", "ERROR").upper()`

And then change the logger from `elasticapm.errors` to `elasticapm` :)

I can also open a PR if that is preferable 

For Python/Flask we need to change a couple things, I'll have a closer look tomorrow

cachedout:
@beniwohli Just circling back around here because of these recent failures:

https://apm-ci.elastic.co/job/apm-integration-tests/job/master/558/

Since you're more familiar with what these various Python environments need, do you think that you could possibly put together a PR to enable debugging logging across the Python tests?

beniwohli:
@cachedout sure, I'll see if I can open a PR later today

PR #895, opened 2020-08-03
chore: move e2e tests from test-clusters to APM ITs

mdelapenya:
## What does this PR do?
This PR adds the foundation to write e2e tests for the environment provisioned by the APM ITs, using Cypress + Cucumber. In this particular PR, we are covering one single path: from an APM transaction to Logs.

It also adds a UI visualisation of test results, using an HTML format.

## Why is it important?
The main goal for these high-level-of-abstraction tests is to cover the most critical user paths for the Observability solution
, specially those paths that are cross-project.

Another benefit of having these tests here is to make it easier for Observability developers to contribute use cases here.

## How to test this
Please start the compose file, as explained in the docs. I.e.

```shell
$ ./scripts/compose.py start --release --with-opbeans-go 7.9.0-SNAPSHOT
```

Then place your terminal in the root directory of these tests and there execute Node's classic:

```shell
$ cd tests/functional
$ npm run test
```

It will spin up Cypress app so you can interact with its UI to run the tests.

On the other hand, if you prefer running the tests in a non-interactive manner, as the CI does, please run:

```shell
$ npm run test:ci
```
To open the HTML visualisation off the test results:
```shell
$ open ./cypress/html-reports/index.html
```

## Follow-ups
We could think about moving the existing e2e tests under the Kibana repo to here, where developers will find it easier to execute in a day-to-day basis. For that, we would need to update the CI job that triggers those tests, simply adding the execution of this new set of tests into the APM ITs pipeline.

### DIFF
--- a/tests/functional/README.md
+++ b/tests/functional/README.md
@@ -1,0 +1,85 @@
+# End-To-End tests for the Observability solution
+
+We want to make sure that any change in the different pieces of the Observability solution satisfy certain critical checks,
+so for that reason we are adding [smoke tests](http://softwaretestingfundamentals.com/smoke-testing/) to verify that the any execution of this test suite includes the critical checks too.
+
+>Smoke Testing, also known as “Build Verification Testing”, is a type of software testing that comprises of a non-exhaustive set of tests that aim at ensuring that the most important functions work. The result of this testing is used to decide if a build is stable enough to proceed with further testing.
+
+## Tooling
+
+The specification of these smoke tests has been done using the `BDD` (Behaviour-Driven Development) principles, where:
+
+>BDD aims to narrow the communication gaps between team members, foster better understanding of the customer and promote continuous communication with real world examples.
+
+The implementation of these smoke tests has been done with [Cypress](https://www.cypress.io/) + [Cucumber](https://cucumber.io/).
+
+### Cucumber: BDD at its core
+
+From their website:
+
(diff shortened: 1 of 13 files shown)


mdelapenya:
I'd like to open a discussion with @elastic/apm-ui and @elastic/logs-metrics-ui teams about the suitability o having these tests here. Please let us know about your thoughts

tests/functional/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+From their website:
+
+>Cucumber is a tool that supports Behaviour-Driven Development(BDD), and it reads executable specifications written in plain text and validates that the software does what those specifications say. The specifications consists of multiple examples, or scenarios.
+
+The way we are going to specify our software, which is the deployment of the test clusters, is using [`Gherkin`](https://cucumber.io/docs/gherkin/reference/).
+
+>Gherkin uses a set of special keywords to give structure and meaning to executable specifications. Each keyword is translated to many spoken languages. Most lines in a Gherkin document start with one of the keywords.
+
+The key part here is **executable specifications**: we will be able to automate the verification of the spefications anf potentially get a coverage of these specs.
```
cachedout, line 26:
```suggestion
The key part here is **executable specifications**: we will be able to automate the verification of the specifications and potentially get a coverage of these specs.
```

mdelapenya:
Closing as obsolete

PR #898, opened 2020-08-17
make log level of elasticapm loggers configurable in flask and django apps

beniwohli:
## What does this PR do?

As logging is configured on an application basis in Python, the agent doesn't have
the capability to configure its own logger (this might change in the future). As
such, we're checking for the ELASTIC_APM_LOG_LEVEL envionment variable in both the
Django and Flask integration test apps.

## Why is it important?

Some other agents support this environment variable. This way we align behavior of
the integration test apps.

## Related issues
See #894

### DIFF
--- a/docker/python/django/testapp/testapp/settings.py
+++ b/docker/python/django/testapp/testapp/settings.py
@@ -114,6 +114,11 @@ LOGGING = {
             'handlers': ['elasticapm'],
             'propagate': False,
         },
+        'elasticapm': {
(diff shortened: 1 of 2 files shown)


kuisathaverat:
docker/python/flask/app.py needs a new line at the end 😄 

```
- hook id: end-of-file-fixer
- exit code: 1
- files were modified by this hook
Fixing docker/python/flask/app.py
```

kuisathaverat (approved):

PR #938, opened 2020-10-12
Send failures to slack channel

cachedout:
## What does this PR do?

Sends failures to the #observerblt-bots channel in Slack

## Why is it important?
Notifies teams of possible issues.

## Related issues
Closes https://github.com/elastic/observability-robots/issues/214

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -32,6 +32,7 @@ pipeline {
   parameters {
     string(name: 'ELASTIC_STACK_VERSION', defaultValue: "8.0.0", description: "Elastic Stack Git branch/tag to use")
     string(name: 'BUILD_OPTS', defaultValue: "", description: "Additional build options to pass to compose.py")
+    string(name: 'SLACK_CHANNEL', defaultValue: 'observablt-bots', description: 'The Slack channel where errors will be posted')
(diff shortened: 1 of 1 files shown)


v1v (approved):

mdelapenya (approved):

.ci/Jenkinsfile
```diff
@@ -123,9 +124,19 @@ pipeline {
     cleanup {
       notifyBuildResult(downstreamJobs: itsDownstreamJobs)
     }
+    unsuccessful {
+      whenFalse(isPR()) {
+        doSlackSend()
+      }
+    }
```
v1v, line 131: You can use

```
	      notifyBuildResult(downstreamJobs: itsDownstreamJobs, slackComment: true, slackChannel: "#${params.SLACK_CHANNEL}")
```

And then the slack message will be sent for broken builds with the below information:

image

v1v (changes requested):
I forgot to mention, if we use:


We get the below slack message for free:

image

cachedout:
@v1v Ah, OK. I had considered that but wasn't sure if there was a reason we weren't *already* using that so I went with the more tailored version. :) I'll make this change. Thanks!

v1v (approved):

PR #944, opened 2020-10-19
fix: change to use adoptOpenJDK to fix the build

kuisathaverat:
## What does this PR do?

It changes to use AdoptOpenJDK Docker image to build the Agent Java test container.

## Why is it important?

For some reason the OpenJDK:10-jre-slim stop to configure libc6:amd64 and this breaks the build. The. AdoptOpenJDK Docker image does not have this issue.

### DIFF
--- a/docker/java/spring/Dockerfile
+++ b/docker/java/spring/Dockerfile
@@ -19,16 +19,14 @@ RUN cd /agent/apm-agent-java \
 COPY maven-package.sh /agent
 RUN ./maven-package.sh "${JAVA_AGENT_BUILT_VERSION}"
 
-FROM openjdk:10-jre-slim
+FROM adoptopenjdk:11-jre-hotspot
 COPY --from=0 /app /app
 COPY --from=0 /agent/apm-agent.jar /app
 RUN apt-get -qq update \
-  && apt-get -qq install -y curl \
+  && apt-get -qq install -y curl --no-install-recommends curl\
   && apt-get -qq clean \
   && rm -fr /var/lib/apt/lists/*
 WORKDIR /app
 EXPOSE 8090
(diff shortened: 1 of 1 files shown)


docker/java/spring/Dockerfile
```diff
@@ -19,16 +19,14 @@ RUN cd /agent/apm-agent-java \
 COPY maven-package.sh /agent
 RUN ./maven-package.sh "${JAVA_AGENT_BUILT_VERSION}"
 
-FROM openjdk:10-jre-slim
+FROM adoptopenjdk:11-jre-hotspot
 COPY --from=0 /app /app
 COPY --from=0 /agent/apm-agent.jar /app
 RUN apt-get -qq update \
-  && apt-get -qq install -y curl \
+  && apt-get -qq install -y curl --no-install-recommends curl\
```
v1v, line 26:
```suggestion
  && apt-get -qq install -y --no-install-recommends curl \ 
```

docker/java/spring/Dockerfile
```diff
@@ -19,16 +19,14 @@ RUN cd /agent/apm-agent-java \
 COPY maven-package.sh /agent
 RUN ./maven-package.sh "${JAVA_AGENT_BUILT_VERSION}"
 
-FROM openjdk:10-jre-slim
+FROM adoptopenjdk:11-jre-hotspot
```
mdelapenya, line 22: I guess this Java version bump is intended, right?

mdelapenya (approved):

docker/java/spring/Dockerfile
kuisathaverat: yep, let's use the same we use in master

v1v (approved):

PR #959, opened 2020-11-04
fix: disable the Elasticsearch dump by default

kuisathaverat:
## What does this PR do?

* Add a default value for the ENABLE_ES_DUMP env var and adapt the condition
* Remove everything not needed from the test Docker container

## Why is it important?

The Elasticsearch dump needs some parameters, without them it shows authentication errors in the console, this can confuse people, so it should be disabled by default. 
The Python Docker container build for run the test has more files it needs, in my local, the Docker context is about 4GB (I have a bunch of folders with data) to build this Docker container when it should be about 30MB, by modifying the .dockerignore file this Docker context takes only the files needed

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -16,5 +16,4 @@ apm-server
 docker-compose.yml
 docker/
 htmlcov/
-scripts/
 venv/
--- a/.dockerignore
+++ b/.dockerignore
@@ -5,11 +5,16 @@
 **/tmp
 *.py[cod]
 *.pyc
-.git
 .cache
+.ci/
 .coverage
+.git
+.github
+.pytest_cache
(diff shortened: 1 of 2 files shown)


mdelapenya (approved):

cachedout (approved):

v1v (approved):

.dockerignore
```diff
@@ -5,11 +5,16 @@
...
 .coverage
+.git
+.github
+.pytest_cache
 /.idea
 apm-server
 docker-compose.yml
+docker/
 htmlcov/
+scripts/
```
kuisathaverat, line 19:
```suggestion
```

PR #962, opened 2020-11-06
Remove ELASTIC_APM_LOG_LEVEL from python tests/apps

basepi:
## What does this PR do?

Remove ELASTIC_APM_LOG_LEVEL from python tests/apps

## Why is it important?

Previously, this didn't actually do anything because log_level
was not an established config value for the python agent.

As of https://github.com/elastic/apm-agent-python/pull/946, this is
now a setting in python and it was causing tests to fail due to the
way python logging works. This PR cleans up all of those references
and will make tests pass.

## Related issues
https://github.com/elastic/apm-agent-python/pull/946


scripts/modules/cli.py
```diff
@@ -363,7 +363,6 @@ def init_start_parser(self, parser, services, argv=None):
             action="store",
             help="APM log level to use",
             choices=["off", "error", "warn", "info", "debug", "trace"],
-            default=DEFAULT_APM_LOG_LEVEL
```
basepi, line 366: All uses of this setting use `self.options.get` with a default,
so we can remove the default in `argparse`. This allows the python
agent to only use this setting if it's explicitly set, which is better
for how the python agent's `log_level` works.

kuisathaverat (approved):

PR #966, opened 2020-11-09
fix: set default log level

kuisathaverat:
## What does this PR do?

It sets the default value for the option `--apm-log-level`

## Why is it important?

https://github.com/elastic/apm-integration-testing/pull/962 causes a regresion due to `self.options.get("apm_log_level", DEFAULT_APM_LOG_LEVEL)` returns `None`instead of the default value, then when we make `self.options.get("apm_log_level", DEFAULT_APM_LOG_LEVEL).lower()` we are making `None.lower()` causing 

```
  File "/Users/inifc/src/apm-integration-testing/scripts/modules/apm_agents.py", line 587, in _content
    self.options.get("apm_log_level", DEFAULT_APM_LOG_LEVEL).lower()
AttributeError: 'NoneType' object has no attribute 'lower'
make[1]: *** [start-env] Error 1
make: *** [env-agent-dotnet] Error 2
```

### DIFF
--- a/scripts/modules/cli.py
+++ b/scripts/modules/cli.py
@@ -13,7 +13,7 @@ import re
 from .beats import BeatMixin
 from .helpers import load_images
 from .opbeans import OpbeansService, OpbeansRum
-from .service import Service, DEFAULT_APM_SERVER_URL
+from .service import Service, DEFAULT_APM_SERVER_URL, DEFAULT_APM_LOG_LEVEL
 
 # these imports are used by discover_services function to discover services from modules loaded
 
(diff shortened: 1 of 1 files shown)


basepi:
Grrr, this is `argparse`'s fault. Unfortunately, this fix will break the Python tests again, because I need to be able to differentiate between the default and when someone explicitly sets the argument.

I'm going to submit an alternative solution in a few minutes, along with another fix for the python tests.

basepi:
Opened #969 which should fix these errors.

Issue #964, opened 2020-11-09
docker-compose-wait does not work if you do not have a clean docker images history clean


kuisathaverat:
docker-compose-wait check the docker images ran in your system and check is all are healthy, this work well when you Docker history (docker ps -a) only have healthy Docker images, but does not work if you have unhealthy Docker images in the history, when this happens the command failed.

```
[ERROR] Failed waiting for all containers are healthy
make[1]: *** [dockerized-test] Error 1
make: *** [docker-test-agent-python] Error 2
```

We have to find a more proper wait to wait for all Docker images ready, I think that a proper solution is to add a docker container that depends on all docker container then this container will wait until all are running.

v1v:
> We have to find a more proper wait to wait for all Docker images ready, I think that a proper solution is to add a docker container that depends on all docker container then this container will wait until all are running.

+100, we use something similar in the opbeans, see https://github.com/elastic/opbeans-java/blob/0ac733323620c2d2d3a0325b2b4fa1bb17601a8d/docker-compose.yml#L117-L121

PR #968, opened 2020-11-09
feta: add a wait container

kuisathaverat:
## What does this PR do?

* Add a new wait container
* Remove docker-compose-wait dependency

## Why is it important?

This new wait container depends on all enabled services, this makes that until all services are not healthy this container is not started. In this way, docker-compose only exits when all containers are healthy or one fails to start.

## Related issues
Closes https://github.com/elastic/apm-integration-testing/issues/964

### DIFF
--- a/scripts/modules/aux_services.py
+++ b/scripts/modules/aux_services.py
@@ -120,7 +120,6 @@ class Zookeeper(Service):
         )
 
 
-
 class WaitService(Service):
     """Create a service that depends on all services ."""
 
--- a/scripts/modules/aux_services.py
+++ b/scripts/modules/aux_services.py
@@ -136,9 +136,6 @@ class WaitService(Service):
             container_name="wait",
             image="busybox",
             depends_on=self.depends_on,
-            healthcheck={
-                "interval": "10s",
-                "retries": "100",
-                "test": ["CMD", "exit 0"],
-            }
+            labels=None,
+            logging=None,
         )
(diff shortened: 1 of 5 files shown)


mdelapenya (approved):

scripts/modules/aux_services.py
```diff
@@ -118,3 +118,27 @@ def _content(self):
             logging=None,
             ports=[self.publish_port(self.port, self.SERVICE_PORT)],
         )
+
+
+
+class WaitService(Service):
+    """Creeate a service that depends on all services ."""
```
v1v, line 125:
```suggestion
    """Create a service that depends on all services ."""
```

scripts/modules/cli.py
```diff
@@ -606,7 +607,7 @@ def start_handler(self):
                 del services["opbeans-load-generator"]
 
         compose = dict(
-            version="2.1",
+            version="2.4",
```
v1v, line 610: I'm not sure if this will cause an issue with the `"condition": "service_healthy"`, as far as I see 2.4 is supported -> https://docs.docker.com/compose/compose-file/compose-file-v2/#depends_on So i said nothing

v1v (approved):

scripts/modules/cli.py
kuisathaverat: no, this is on 3.0

cachedout (approved):

Issue #661, opened 2019-11-01
Add an option to build Kibana from source


jalvz:
Similar to the `--apm-server-build` flag.
Ideally, would be able to take a remote/branch param too

POC dockerfile:
```dockerfile
FROM node:10

RUN git clone git@github.com:elastic/kibana.git \
 cd kibana/ \
 yarn kbn bootstrap \

CMD  yarn start
```

graphaelli:
Took a quick a look at this and see some challenges:

* need to continue using something like `/usr/local/bin/kibana-docker` to allow passing configuration via env vars
* probably need to continue using `/usr/local/bin/dumb-init` as entrypoint as kibana-docker doesn't handle signals properly
* node version has to match exactly, so need `node:10.15.2` today, but could use a param

Kibana's Dockerfile is generated from https://github.com/elastic/kibana/blob/master/src/dev/build/tasks/os_packages/docker_generator/templates/dockerfile.template.js, might be useful when this is picked up.

kuisathaverat:
I am digging a little on it there is a command to build the Kibana Docker container, so I am going to try to spin up a container that builds the Kibana Docker container then we use the Docker container generated.

kuisathaverat:
To compile Kibana is a pain 😞 

```
[18:33:57] Finished 'build' after 18 min
   │ERROR failure
   │ERROR NestedError: Cannot copy from `/kibana/x-pack/build/plugin/kibana/x-pack/NOTICE.txt` to `/kibana/build/kibana/x-pack/NOTICE.txt`: utimes `/kibana/build/kibana/x-pack/NOTICE.txt` failed: EMFILE: too many open files, utime '/kibana/build/kibana/x-pack/NOTICE.txt'
   │          at cpFile.on.then.catch.error (/kibana/packages/kbn-pm/dist/index.js:80158:13)
   │      Caused By: NestedError: utimes `/kibana/build/kibana/x-pack/NOTICE.txt` failed: EMFILE: too many open files, utime '/kibana/build/kibana/x-pack/NOTICE.txt'
   │          at fsP.utimes.catch.error (/kibana/packages/kbn-pm/dist/index.js:105810:8)
   │      Caused By: Error: EMFILE: too many open files, utime '/kibana/build/kibana/x-pack/NOTICE.txt'

/kibana/node_modules/@babel/register/lib/cache.js:80
        throw e;
        ^

[...]

PR #970, opened 2020-11-09
feat: build kibana container from sources

kuisathaverat:
## What does this PR do?

It adds a new flag to build Kibana from sources in a folder.

## Why is it important?

Now developers can run their Kibana code directly against the Elastic Stack using compose.py.

There are two new flags `--kibana-src` and `kibana-src-start-cmd` 
* `--kibana-src` is the absolute path to a local folder with the kibana sources.
* `kibana-src-start-cmd` it replaces the default command used to start kibana from sources `yarn kbn bootstrap && yarn start -c /usr/share/kibana/config/kibana_src.yml -c /usr/share/kibana/config/kibana.yml --no-dev-config`.

The default starts command loads two configuration files `/usr/share/kibana/config/kibana_src.yml` that configure the docker-compose Elasticsearch and `/usr/share/kibana/config/kibana.yml` that is the file inside `config/kibana.yml` in the Kibana sources folder.

## Limitations 

It only works with the Elasticsearch instance deploy in the Docker compose, to use other Elasticsearch instance you have to use `--kibana-src-start-cmd` to replace the default configurations.
All configuration made on environment variables is ignored, so the only way to configure the settings is by using the `config/kibana.yml` file.

## Local testing

```shell
git clone git@github.com:elastic/kibana.git
cd kibana
git checkout 7.13
export KIBANA_SRC_DIR=$(pwd)
cd -
git clone git@github.com:kuisathaverat/apm-integration-testing.git
cd apm-integration-testing
git checkout -b origin/kibana_src kibana_src
./scripts/compose.py start 7.13.4 --kibana-src=${KIBANA_SRC_DIR} --no-apm-server
```

## Related issues
Closes https://github.com/elastic/apm-integration-testing/issues/661

### DIFF
--- a/scripts/modules/elastic_stack.py
+++ b/scripts/modules/elastic_stack.py
@@ -1167,8 +1167,14 @@ class Kibana(StackService, Service):
         parser.add_argument(
             "--kibana-src-start-cmd",
             nargs="?",
-            help="Command used to start Kibana from sources (yarn kbn bootstrap && yarn start).",
-            default="yarn kbn bootstrap && yarn start"
+            help="Command used to start Kibana from sources " +
+                 "(yarn kbn bootstrap && yarn start " +
+                 "-c /usr/share/kibana/config/kibana_src.yml " +
+                 "-c /usr/share/kibana/config/kibana.yml).",
+            default="yarn kbn bootstrap && yarn start " +
+                    "-c /usr/share/kibana/config/kibana_src.yml " +
+                    "-c /usr/share/kibana/config/kibana.yml " +
+                    "--no-dev-config"
         )
 
     def _content(self):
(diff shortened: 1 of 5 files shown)


cachedout (approved):

scripts/modules/elastic_stack.py
```diff
@@ -844,6 +853,12 @@ def add_arguments(cls, parser):
             help="disable the APM service maps UI",
         )
 
+        parser.add_argument(
+            "--kibana-from-sources",
```
mdelapenya, line 857: wdyt about renaming the flag to something easier to recall, like `--kibana-src`?

mdelapenya (approved):

scripts/modules/elastic_stack.py
kuisathaverat, line 857:
```suggestion
            "--kibana-src",
```

Issue #981, opened 2020-11-16
7.4 --release does not work


v1v:
### Stacktrace

```
Starting stack services..

Pulling elasticsearch ... error
Pulling apm-server    ... error
Pulling wait-service  ... done

ERROR: for elasticsearch  manifest for docker.elastic.co/elasticsearch/elasticsearch:7.4.3 not found: manifest unknown: manifest unknown

ERROR: for apm-server  manifest for docker.elastic.co/apm/apm-server:7.4.3 not found: manifest unknown: manifest unknown
ERROR: manifest for docker.elastic.co/elasticsearch/elasticsearch:7.4.3 not found: manifest unknown: manifest unknown
manifest for docker.elastic.co/apm/apm-server:7.4.3 not found: manifest unknown: manifest unknown
Traceback (most recent call last):
  File "/Users/vmartinez/work/src/github.com/elastic/apm-integration-testing/scripts/compose.py", line 19, in <module>
    main()
  File "/Users/vmartinez/work/src/github.com/elastic/apm-integration-testing/scripts/compose.py", line 15, in main
    setup()
  File "/Users/vmartinez/work/src/github.com/elastic/apm-integration-testing/scripts/modules/cli.py", line 184, in __call__
    self.args.func()
  File "/Users/vmartinez/work/src/github.com/elastic/apm-integration-testing/scripts/modules/cli.py", line 663, in start_handler
    self.run_docker_compose_process(docker_compose_cmd + pull_params + image_services)
  File "/Users/vmartinez/work/src/github.com/elastic/apm-integration-testing/scripts/modules/cli.py", line 433, in run_docker_compose_process
    subprocess.check_call(docker_compose_cmd)
  File "/usr/local/Cellar/python@3.9/3.9.0_1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/subprocess.py", line 373, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['docker-compose', '-f', '/Users/vmartinez/work/src/github.com/elastic/apm-integration-testing/docker-compose.yml', 'pull', 'elasticsearch', 'wait-service', 'apm-server']' returned non-zero exit status 1.

```

### How to reproduce locally

```bash
$ python3 scripts/compose.py start 7.4 \
                  --release \
                  --build-parallel --no-kibana  \
                  --force-build
```

PR #982, opened 2020-11-16
Fix 7.4.2 to 7.4

v1v:
## What

7.4.3 never was released

## Issues

Closes https://github.com/elastic/apm-integration-testing/issues/981


mdelapenya (approved):

scripts/modules/cli.py
```diff
@@ -70,7 +70,7 @@ class LocalSetup(object):
         '7.1': '7.1.1',
         '7.2': '7.2.1',
         '7.3': '7.3.2',
-        '7.4': '7.4.3',
+        '7.4': '7.4.2',
```
v1v, line 73: See https://github.com/elastic/apm-integration-testing/blob/e0de7b66fd636b5f382a915b00ded9fb06967761/scripts/compose.py#L2260

PR #975, opened 2020-11-12
fix: retry maven download

kuisathaverat:
## What does this PR do?

it increases the retries to 10

## Why is it important?

Sometimes seems like 3 are not enough 

```
[2020-11-12T14:11:57.360Z] [ERROR] Failed to execute goal org.apache.maven.plugins:maven-enforcer-plugin:3.0.0-M1:enforce (enforce-java) on project apm-agent-parent: Execution enforce-java of goal org.apache.maven.plugins:maven-enforcer-plugin:3.0.0-M1:enforce failed: Plugin org.apache.maven.plugins:maven-enforcer-plugin:3.0.0-M1 or one of its dependencies could not be resolved: Failed to collect dependencies at org.apache.maven.plugins:maven-enforcer-plugin:jar:3.0.0-M1 -> org.apache.maven.enforcer:enforcer-rules:jar:3.0.0-M1 -> org.apache.maven.shared:maven-common-artifact-filters:jar:3.0.1: Failed to read artifact descriptor for org.apache.maven.shared:maven-common-artifact-filters:jar:3.0.1: Could not transfer artifact org.apache.maven.shared:maven-common-artifact-filters:pom:3.0.1 from/to central (https://repo.maven.apache.org/maven2): Transfer failed for https://repo.maven.apache.org/maven2/org/apache/maven/shared/maven-common-artifact-filters/3.0.1/maven-common-artifact-filters-3.0.1.pom: readHandshakeRecord: Connection reset by peer (Write failed) -> [Help 1]
[2020-11-12T14:11:57.361Z] [ERROR] 
[2020-11-12T14:11:57.361Z] [ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[2020-11-12T14:11:57.361Z] [ERROR] Re-run Maven using the -X switch to enable full debug logging.
[2020-11-12T14:11:57.361Z] [ERROR] 
[2020-11-12T14:11:57.361Z] [ERROR] For more information about the errors and possible solutions, please read the following articles:
[2020-11-12T14:11:57.361Z] [ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/PluginResolutionException
```

```
[2020-11-12T14:19:30.026Z] ERROR: for agent-java-spring  (<Service: agent-java-spring>, 'The command \'/bin/sh -c ./maven-package.sh "${JAVA_AGENT_BUILT_VERSION}"\' returned a non-zero code: 1')
[...]

PR #983, opened 2020-11-16
Use cached .m2 dependencies

v1v:
## What does this PR do?

Opinionated approach to copy the .m2 folder in the docker image to run the maven builds using the cached m2 repo from the CI workers.

For such if the `--java-m2-cache` is enabled the entrypoint will set `MAVEN_CONFIG='-Dmaven.repo.local=/root/.m2/repository'` and also coNfigure the mvn calls with `-Dmaven.repo.local=/root/.m2/repository`.

If you run this locally, then. you need to copy your `.m2/repository` folder to the `docker/java/spring/.m2`.

In the CI, all the scripts are now doing that particular copy, see https://github.com/elastic/apm-integration-testing/pull/983/files#diff-155c3177c3fc74ae3606124ec237271e50b69821af6a1c1232bc3da9ab387c37R17-R20

## Why is it important?

Reduce the connection resets errors in the builds

## Related issues
Supersedes https://github.com/elastic/apm-integration-testing/pull/975

## Tasks

- [x] Add UTs


docker/java/spring/Dockerfile
```diff
@@ -17,6 +18,7 @@ RUN cd /agent/apm-agent-java \
   && git checkout ${JAVA_AGENT_BRANCH}
 
 COPY maven-package.sh /agent
+COPY .m2 /.m2
```
kuisathaverat, line 21: if the folder does not exist it fails

v1v: I missed to copy the folder. Besides, what do you think? is this a sensible approach?

cachedout (approved):

mdelapenya (approved):

felixbarny (approved):
I'm looking forward to cached dependencies in the Java agent and ecs-logging Java! 🙂

.gitignore
```diff
@@ -30,3 +30,6 @@ target/
 
 # For the docs generation
 .apm-its
+
+# For the cache java dependencies
+docker/java/spring/.m2
```
SylvainJuge, line 35: Why do we have `spring` in the folder name here ? It seems to be a generic maven repository, not really related to spring-framework.

v1v: .m2 folder empty folder was created in the repo to enable this particular feature, the scripts that run in the CI copy their `~/.m2` folder to `docker/java/spring`, other than that I don't know what you mean :?

SylvainJuge: I mean that not all dependencies are spring dependencies within a local maven repository, thus this `spring` folder will contain things that are not related to `spring` at all, thus the `spring` here is definitely confusing, using `docker/java/maven` instead of `docker/java/spring` would make more sense.

v1v: I see, no strong opinion, neither I know the reason for that particular folder name, so I'd say that folder name could be renamed in another follow up

kuisathaverat: the folder `docker/java/spring` has this name because it is a spring-boot application, change the name to `docker/java/maven` does not have any sense.

kuisathaverat (approved):

PR #984, opened 2020-11-18
add enterprise search

graphaelli:
## What does this PR do?

Adds enterprise search into apm-integration-testing using `--with-enterprise-search` option.  Also adds a convenience command for resetting the enterprise search user password using `compose.py reset-enterprise-search-password`.

## Why is it important?

Enterprise search has the Ruby agent integrated.  By default APM is enabled:

image


kovyrin:
If you provision Enterprise Search with `ENT_SEARCH_DEFAULT_PASSWORD=something`, it will automatically pick that up and use it as the password for the default user. Makes it a lot easier to write all kinds of testing automation for it.

graphaelli:
Thanks @kovyrin , will push up a change including that.

kuisathaverat (approved):

v1v (approved):

Issue #986, opened 2020-11-19
wait-service  Service "heartbeat" is missing a healthcheck configuration


v1v:
I pulled the latest version of the APM-integration-test env, and am getting an error starting up.  First I had to update my version of docker-compose because the compose version was updated to 2.4.
Now I am running into an issue with "wait-service" :
`ERROR: for wait-service  Service "heartbeat" is missing a healthcheck configuration`

It looks like it's trying to add a wait-service for every container:

```
    "wait-service": {
      "container_name": "wait",
      "depends_on": {
        "apm-server": {
          "condition": "service_healthy"
        },
        "elasticsearch": {
          "condition": "service_healthy"
        },
        "filebeat": {
          "condition": "service_healthy"
        },
        "heartbeat": {
          "condition": "service_healthy"
        },
(...)
```
But none of the *beats have health checks implemented.  Looking at the code I can't see where it's generating that `wait-service` block.  Does anyone know where comes from?  

### Regression

Most likely https://github.com/elastic/apm-integration-testing/issues/964

### Workaround

- Removing the sections from my docker-compose manually starts fine

PR #990, opened 2020-11-19
Fix services without healthcheck

v1v:
## What does this PR do?

Avoid adding healtcheck for services that don't have any healthcheck.

## Why is it important?

Fix a regression

## Related issues
Closes https://github.com/elastic/apm-integration-testing/issues/986

### DIFF
--- a/scripts/modules/aux_services.py
+++ b/scripts/modules/aux_services.py
@@ -140,4 +140,10 @@ class WaitService(Service):
         )
 
     def isHealthCheckServiceDisabled(self, name):
-        return name == "opbeans-load-generator" or name == "filebeat" or name == "heartbeat" or name == "metricbeat" or name == "packetbeat"
+        return (
+            name == "opbeans-load-generator" or
+            name == "filebeat" or
+            name == "heartbeat" or
+            name == "metricbeat" or
+            name == "packetbeat"
+        )
--- a/scripts/modules/aux_services.py
+++ b/scripts/modules/aux_services.py
@@ -129,7 +129,7 @@ class WaitService(Service):
 
     def _content(self):
         for s in self.services:
-            if s.name() != self.name() and s.name() != "opbeans-load-generator":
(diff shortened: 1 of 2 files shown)


scripts/modules/aux_services.py
```diff
@@ -138,3 +138,12 @@ def _content(self):
             labels=None,
             logging=None,
         )
+
+    def isHealthCheckServiceDisabled(self, name):
+        return (
+            name == "opbeans-load-generator" or
+            name == "filebeat" or
```
graphaelli, line 145: just thought, but how about enabling [`http.enabled`](https://www.elastic.co/guide/en/beats/filebeat/master/http-endpoint.html) and healthchecking that for all beats?

v1v: Sure, let me raise a new PR with that particular check :)

scripts/modules/aux_services.py
```diff
@@ -138,3 +138,12 @@ def _content(self):
             labels=None,
             logging=None,
         )
+
+    def isHealthCheckServiceDisabled(self, name):
+        return (
```
cachedout, line 143: Nit: Could this be rewritten as `return name in ["opbeans-load-generator" , "filebeat", ... ]` and still achieve what you are looking for? It might read more cleanly.

cachedout (approved):

mdelapenya (approved):

scripts/modules/aux_services.py
v1v: 💯

scripts/modules/aux_services.py
kuisathaverat: yep, we use it for the DoD Docker images https://repo1.dsop.io/dsop/elastic/beats/auditbeat/-/blob/development/Dockerfile#L90

kuisathaverat (approved):

v1v:
Since we got https://github.com/elastic/apm-integration-testing/pull/991 in place, I thought it might take us a bit longer, I'll proceed to close this PR for now

Issue #1004, opened 2020-12-08
Add Elastic Agent container


jalvz:
The default policy should start apm-server.

Critically, this requires a service that installs the APM package prior to any data ingestion.

graphaelli:
I think this should depend on the version of the stack.  One idea:

For any version:
  -  `--with-elastic-agent` - plain old agent

For >= 8.0:
  -  `--with-apm-server` - agent with automatically configured apm-server. 

For < 8.0:
  -  `--with-apm-server` -  no change, apm-server container started
  -  `--with-apm-server-agent` - agent with automatically configured apm-server.  Looking ahead to implementation, this should probably automatically set `--apm-server-url` but container dependencies are going to be tricky.

jalvz:
Sounds good to me, except first bullet point - we only have agent/apm-server combo starting 7.11, so what is it for?

graphaelli:
That would be for non-apm integrations

graphaelli:
#1005 added `--with-elastic-agent`.

To automatically install the APM Integration this needs automation:

```
curl 'http://kibana:5601/api/fleet/package_policies' -H 'Content-Type: application/json' --data-raw '{"name":"apm-1","description":"","namespace":"default","policy_id":"27f16170-3b38-11eb-af1b-f5e1393cd644","enabled":true,"output_id":"","inputs":[{"type":"apm","enabled":true,"streams":[],"vars":{"enable_rum":{"type":"bool","value":true}}}],"package":{"name":"apm","title":"Elastic APM","version":"0.1.0-dev.1"}}'
```

where `policy_id` and `package.version` are variable.

@jalvz :
1. Is there a reason not to attach the integration to the the default policy? 
2. How often does the package version change?

It seems we can look up the default policy in a number of ways.  `/api/fleet/agent_policies` appears to be the most proper but I haven't verified whether that has filtering params or if they must all be paginated through looking for `is_default: true`.  

Querying the global fleet settings also seems to do it and require no pagination but seems like a hack -
```
GET http://kibana:5601/api/fleet/settings
{"item":{"id":"27ee0610-3b38-11eb-af1b-f5e1393cd644","agent_auto_upgrade":true,"package_auto_upgrade":true,"kibana_urls":["http://kibana:5601"]}}
```

jalvz:
[...]

PR #1005, opened 2020-12-11
Add Elastic Agent

graphaelli:
## What does this PR do?

Adds Elastic Agent into apm-integration-testing using `--with-elastic-agent` option.  Elastic Agent is configured in fleet mode with system logs and metrics integrations, the usual defaults.

## Why is it important?

For testing integrations with Elastic Agent eg APM.

Some things to note:

* `elastic-agent.yml` can't be bind mounted into the container as agent tries to move it out of the way when fleeting.  As a result it seems any customizations will have to happen after the agent is running - this will be follow as a separate effort for #1004.
* The kibana instance in apm-it will direct agents to use it for fleet along with the configured elasticsearch - support for using a different kibana seems unimportant right now
* The setup and enroll commands aren't silent so docker logs for this container are not nice.  This message seems to just be a warning: `Elastic Agent might not be running; unable to trigger restart`
* Elastic agent logs to various files within the container and not to stdout like usual docker logging - it ships its own logs so that's nice but it's surprising while debugging unless you know this is happening
* The APM integration is  listed but can't be installed, currently showing the following.  This does not block this pr:
image

## Related issues
related #1004


simitt:
Great progress!

> The APM integration is listed but can't be installed, currently showing the following. This does not block this pr:

I managed to install the APM integration via navigating to _Policies/Add integration_. 
Screenshot 2020-12-11 at 10 40 19

Something is odd when downloading the packages: on the first attempt it always returns a `404` for the `0.1.0-dev1` package, on second try it finds the APM package and downloads it successfully. The package was added by @jalvz yesterday, I don't see why it shouldn't be found on first try. 

The APM Server is not yet bundled nor available via the artifacts API. While I managed to install the APM integration, the APM Server cannot actually start as it is not yet available.

simitt:
The agent could successfully enroll, also be assigned to a new policy; but when installing the APM integration to the policy, the agent went `offline` soon, as it couldn't successfully checkin with fleet; resulting in following log error:
``` 
{"log.level":"error","@timestamp":"2020-12-11T10:10:45.187Z","log.origin":{"file.name":"application/fleet_gateway.go","file.line":176},"message":"Could not communicate with Checking API will retry, error: fail to checkin to fleet: Post \"http://0.0.0.0:5601/api/fleet/agents/d216e510-3b8f-11eb-b13c-0ddb183c8890/checkin?\": dial tcp 0.0.0.0:5601: connect: connection refused","ecs.version":"1.6.0"}
```
Not an issue with this PR, just leaving the comment here in case others try it out too - needs to be fixed separately.

simitt:
In another attempt, starting with completely fresh docker images - the APM Server integration was now successfully installed, without any further changes. 

(0) Starting integration testing with your branch and elastic-agent enabled
(1) Creating a new policy and assigning policy to agent
(2) Adding APM to the policy 
Error on first attempt to check APM integration, and on first attempt to apply changes:
Screenshot 2020-12-11 at 11 36 09
Screenshot 2020-12-11 at 11 37 00
Succesful on second attempt (similar as mentioned in previous comment):
Screenshot 2020-12-11 at 11 37 23

APM Server is now running, and can be reached from inside the elastic agent docker container (port is not opened yet)
Screenshot 2020-12-11 at 11 42 14
Screenshot 2020-12-11 at 11 43 32

graphaelli:
Nice @simitt!  I haven't been able to get it to work adding the integration via a different flow:

APM Integration

I will try the method you suggested.  Any changes need to this pull request?

graphaelli:
I was able to reproduce your result by installing the integration from the policy page:

image

scripts/modules/elastic_stack.py
```diff
@@ -587,6 +587,68 @@ def render_tee(self):
...
+        # build environment
+        # Environment variables used
+        # FLEET_ENROLLMENT_TOKEN - existing enrollment token to be used for enroll
+        # FLEET_ENROLL - if set to 1 enroll will be performed
+        # FLEET_ENROLL_INSECURE - if set to 1, agent will enroll with fleet using --insecure flag
+        # FLEET_SETUP - if  set to 1 fleet setup will be performed
+        # FLEET_TOKEN_NAME - token name for a token to be created
+        # KIBANA_HOST - actual kibana host [http://localhost:5601]
+        # KIBANA_PASSWORD - password for accessing kibana API [changeme]
+        # KIBANA_USERNAME - username for accessing kibana API [elastic]
```
simitt, line 610: isn't that `admin` instead of `elastic`?

scripts/modules/elastic_stack.py
```diff
@@ -587,6 +587,68 @@ def render_tee(self):
...
+            ]
+        )
+
+    @classmethod
+    def add_arguments(cls, parser):
+        super(ElasticAgent, cls).add_arguments(parser)
+        parser.add_argument(
+            "--elastic-agent-kibana-url",
+            default="http://admin:changeme@" + cls.DEFAULT_KIBANA_HOST,
+            help="Elastic Agent's Kibana URL"
```
simitt, line 648: maybe add `including username + password`

simitt (approved):
Great to get this in! Just left minor comments - feel free to skip.

jalvz:
Thanks @graphaelli and @simitt ! 
I could did reproduce the same outcome as you....

scripts/modules/elastic_stack.py
```diff
@@ -859,6 +922,10 @@ def __init__(self, **options):
             if self.at_least_version("7.7"):
                 self.environment["XPACK_SECURITY_ENCRYPTIONKEY"] = "fhjskloppd678ehkdfdlliverpoolfcr"
                 self.environment["XPACK_ENCRYPTEDSAVEDOBJECTS_ENCRYPTIONKEY"] = "fhjskloppd678ehkdfdlliverpoolfcr"
+            if self.at_least_version("7.8"):
+                self.environment["XPACK_FLEET_AGENTS_ELASTICSEARCH_HOST"] = urls[0]
```
jalvz, line 926: The Elastic Agent container crashes in 7.8, and I suspect it might be because these settings were named differently. I just can't find it, https://www.elastic.co/guide/en/kibana/7.10/fleet-settings-kb.html doesn't exist for previous versions...

We might just as well support >=7.10 only and revisit if needed for earlier version?

The error is pretty clear, thou:

```
localtesting_7.8.1_elastic-agent | {"statusCode":404,"error":"Not Found","message":"Not Found"}{"statusCode":404,"error":"Not Found","message":"Not Found"}{"statusCode":404,"error":"Not Found","message":"Not Found"}
```

graphaelli: Sounds good to me, perhaps @simitt can include it as part of the next chunk of #1004

PR #1006, opened 2020-12-16
Bump 7.10.1 release

v1v:
## What does this PR do?

Bump 7.<number> to 7.10.1 version


tests/versions/apm_server.yml
```diff
@@ -1,6 +1,7 @@
 APM_SERVER:
   - '7.x'
   - '7.10'
+  - '7.10;--release'
```
v1v, line 4: I'm not sure whether this is correct :/

kuisathaverat: `--release` is correct, we changed it with the first release 'couse is no longer a SNAPSHOT

tests/versions/apm_server.yml
```diff
@@ -1,6 +1,7 @@
 APM_SERVER:
   - '7.x'
   - '7.10'
```
kuisathaverat, line 3:
```suggestion
```

kuisathaverat (approved):

PR #1008, opened 2020-12-16
[java] Build with another retry if something bad happens

v1v:
## What does this PR do?

As long as we suffer the connection reset when fetching artifacts from the maven central we need to add some workaround to add more resilience to our builds, let's try this other approach


kuisathaverat (approved):

v1v:
jenkins run the tests please

PR #1010, opened 2020-12-18
Add package-registry container

jalvz:
## What does this PR do?

This adds a package registry container with options to run any registry distribution and mount a local APM package.

## Related issues

Part of #1004

### DIFF
--- a/scripts/modules/elastic_stack.py
+++ b/scripts/modules/elastic_stack.py
@@ -596,10 +596,10 @@ class PackageRegistry(StackService, Service):
 
     def __init__(self, **options):
         super(PackageRegistry, self).__init__(**options)
-        self.distribution = options.get("package_registry_distribution", "snapshot")
-        self.apm_package = options.get("apm_package")
         if not self.at_least_version("7.10"):
-            raise Exception("Package registry only supported for 7.70+")
+            raise Exception("Package registry only supported for 7.10+")
+        self.distribution = options.get("package_registry_distribution", "snapshot")
+        self.apm_package = options.get("package_registry_apm_path")
         self.environment = {}
 
     def _content(self):
@@ -610,7 +610,7 @@ class PackageRegistry(StackService, Service):
             ports=[self.publish_port(self.SERVICE_PORT)]
         )
         if self.apm_package:
-            content.update(volumes=[self.apm_package + ":/packages/" + self.distribution + "/apm"])
+            content["volumes"] = [self.apm_package + ":/packages/" + self.distribution + "/apm"]
         return content
 
     @classmethod
@@ -619,10 +619,11 @@ class PackageRegistry(StackService, Service):
         parser.add_argument(
             "--package-registry-distribution",
             default="snapshot",
-            help="Package registry distribution: snapshot | staging | production"
+            choices=['snapshot', 'staging', 'production'],
+            help="Package registry distribution"
         )
         parser.add_argument(
-            "--apm-package",
+            "--package-registry-apm-path",
             help="Folder of a local APM package to add to the registry"
         )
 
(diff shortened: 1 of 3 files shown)


scripts/modules/elastic_stack.py
```diff
@@ -587,6 +587,46 @@ def render_tee(self):
...
+            content.update(volumes=[self.apm_package + ":/packages/" + self.distribution + "/apm"])
+        return content
+
+    @classmethod
+    def add_arguments(cls, parser):
+        super(PackageRegistry, cls).add_arguments(parser)
+        parser.add_argument(
+            "--package-registry-distribution",
+            default="snapshot",
+            help="Package registry distribution: snapshot | staging | production"
```
graphaelli, line 622: are these the only valid values?  If so I'd suggest adding [choices](https://docs.python.org/3/library/argparse.html#choices).

scripts/modules/elastic_stack.py
```diff
@@ -587,6 +587,46 @@ def render_tee(self):
...
+    SERVICE_PORT = "8080"
+
+    docker_path = "package-registry"
+    docker_name = "distribution"
+
+    def __init__(self, **options):
+        super(PackageRegistry, self).__init__(**options)
+        self.distribution = options.get("package_registry_distribution", "snapshot")
+        self.apm_package = options.get("apm_package")
+        if not self.at_least_version("7.10"):
```
graphaelli, line 601: might as well do this first thing

scripts/modules/elastic_stack.py
```diff
@@ -587,6 +587,46 @@ def render_tee(self):
...
+
+    def _content(self):
+        content = dict(
+            image="/".join((self.docker_registry, self.docker_path, self.docker_name)) + ":" + self.distribution,
+            environment=self.environment,
+            healthcheck=curl_healthcheck(self.SERVICE_PORT, path="/", interval="5s", retries=10),
+            ports=[self.publish_port(self.SERVICE_PORT)]
+        )
+        if self.apm_package:
+            content.update(volumes=[self.apm_package + ":/packages/" + self.distribution + "/apm"])
```
graphaelli, line 613: I think this would be less surprising as:

```suggestion
            content["volumes"] = [self.apm_package + ":/packages/" + self.distribution + "/apm"]
```

scripts/modules/elastic_stack.py
```diff
@@ -587,6 +587,46 @@ def render_tee(self):
...
+    @classmethod
+    def add_arguments(cls, parser):
+        super(PackageRegistry, cls).add_arguments(parser)
+        parser.add_argument(
+            "--package-registry-distribution",
+            default="snapshot",
+            help="Package registry distribution: snapshot | staging | production"
+        )
+        parser.add_argument(
+            "--apm-package",
```
graphaelli, line 625: In keeping with the naming of other options:

```suggestion
            "--package-registry-apm-path",
```

scripts/modules/elastic_stack.py
```diff
@@ -587,6 +587,46 @@ def render_tee(self):
...
+
+    docker_path = "package-registry"
+    docker_name = "distribution"
+
+    def __init__(self, **options):
+        super(PackageRegistry, self).__init__(**options)
+        self.distribution = options.get("package_registry_distribution", "snapshot")
+        self.apm_package = options.get("apm_package")
+        if not self.at_least_version("7.10"):
+            raise Exception("Package registry only supported for 7.70+")
```
graphaelli, line 602:
```suggestion
            raise Exception("Package registry only supported for 7.10+")
```

scripts/modules/elastic_stack.py
```diff
@@ -587,6 +587,46 @@ def render_tee(self):
...
+        self.environment = {}
+
+    def _content(self):
+        content = dict(
+            image="/".join((self.docker_registry, self.docker_path, self.docker_name)) + ":" + self.distribution,
```
graphaelli, line 607: can this be:
```suggestion
            image=self.default_image(version_override=self.distribution),
```

graphaelli (changes requested):
great addition, some mostly superficial feedback

scripts/modules/elastic_stack.py
jalvz: This won't work because the package-registry images don't follow conventions, so all those `-SNAPSHOT`, `-oss` suffixes are not there.

Implemented the rest, thanks for the review!

graphaelli: Thanks for checking

graphaelli (approved):

PR #1012, opened 2020-12-21
set kibana SERVER_HOST

graphaelli:
## What does this PR do?

work around issue `"host" must be a valid hostname` with kibana container default `server.host: 0` setting

## Why is it important?

kibana container can't start

related to: https://github.com/elastic/kibana/issues/86716


astorm (approved):
👍

PR #1016, opened 2020-12-31
[elastic-agent]: Support running APM Server managed

simitt:
## What does this PR do?
Supports running APM Server managed by Elastic Agent by running `--apm-server-managed --with-elastic-agent`

* remove existing APM packages and install apm-integration-testing policy to default policy

* expose APM Server port in Elastic Agent

* run a reverse proxy for `apm-server:8200` to not require changes in existing Opbeans services

* only uses config options that are supported by managed APM Server: `apm-server.secret_token`
  (RUM is always enabled, host is set to listen on all interfaces anyways)

The APM Server container runs the install script and provides the healthcheck for APM. This way the Elastic Agent (for now) doesn't have to bother about which integrations should be running for providing an overall healthcheck.

## Why is it important?
To test APM Server managed by Elastic Agent

## Related issues
related https://github.com/elastic/apm-integration-testing/issues/1004

cc @graphaelli I see this as a short-to-midterm solution until we have a more general solution on how managed policies should be created on Elastic Agent startup. There is also going to be a couple of changes when the Fleet Server gets integrated into the Agent, as there probably won't be a direct Kibana connection anymore .


docker/apm-server/managed/docker-entrypoint.sh
```diff
@@ -0,0 +1,72 @@
...
+    exit -2
+fi
+
+# create apm-integration-testing package policy
+# and aadd it to the default agent policy
+pkg_policy_data()
+{
+  cat <<EOF
+{
+  "name":"apm-integration-testing",
```
ruflin, line 41: @nchaulet I'm wondering how much of these json parts are optional and things would still work? Can we skip `output_id`, `namespace`, `streams` any many other things and the correct defaults will be filled in either by KB or Elastic Agent? This would make the API simpler.

graphaelli:
> short-to-midterm solution until we have a more general solution on how managed policies should be created on Elastic Agent startup. There is also going to be a couple of changes when the Fleet Server gets integrated into the Agent, as there probably won't be a direct Kibana connection anymore .

Given this caveat I'm comfortable with approach.

docker/apm-server/managed/docker-entrypoint.sh
```diff
@@ -0,0 +1,72 @@
...
+
+set -e
+
+echo ${KIBANA_HOST}
+
+fleet="${KIBANA_HOST}/api/fleet"
+pkg_policy_name="apm-integration-testing"
+
+# remove existing apm-integration-testing package policy
+pkg_policy_id="$(curl -s -H 'kbn-xsrf: true' -H 'Content-Type: application/json' "${fleet}/package_policies?kuery=ingest-package-policies.name:${pkg_policy_name}" | jq '.items[0].id')"
```
graphaelli, line 11: Why is deletion necessary, is it created other than by this entrypoint?  Can we instead only create if it doesn't already exist?

docker/apm-server/managed/docker-entrypoint.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
+
+set -e
```
graphaelli, line 3: since we're relying on piping to `jq` we should use `-o pipefail`.  If we're going to use this longer term I'd suggest writing this not in bash

scripts/modules/elastic_stack.py
```diff
@@ -24,8 +29,32 @@ class ApmServer(StackService, Service):
...
     def __init__(self, **options):
         super(ApmServer, self).__init__(**options)
-        default_apm_server_creds = {"username": "apm_server_user", "password": "changeme"}
 
+        default_apm_server_creds = {"username": "apm_server_user", "password": "changeme"}
+        self.managed = False
+
+        # run apm-server managed by elastic-agent
+        if self.options.get("apm_server_managed"):
+            self.managed = True
```
graphaelli, line 38: this should wait for kibana if apm-it is also providing kibana

graphaelli (changes requested):

docker/apm-server/managed/docker-entrypoint.sh
nchaulet: Right now all of these fields are mandatory in the API but I agree we can simplify the API and have optional field here

scripts/modules/elastic_stack.py
simitt: I am specifying _depends_on_kibana_ https://github.com/elastic/apm-integration-testing/pull/1016/files#diff-df11512a4602ab3dcd2e68f8f85f7462684d54050bfcb70ba7f94b39d53bea8aR606

graphaelli: gotcha, then I suggest moving the `depends_on` determination logic into `__init__` , via `self.depends_on`

docker/apm-server/managed/docker-entrypoint.sh
simitt: switched to Go implementation, but still deleting any APM packages - the configuration options might not be the same (currently only the secret token).

docker/apm-server/managed/main.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+	return client.deletePackagePolicies(ids)
+}
+
+func addAPMPolicy(client *kibanaClient) error {
+	agentPolicies, err := client.getAgentPolicies("kuery=ingest-agent-policies.is_default:true")
+	if err != nil {
+		return errors.Wrap(err, "addAPMPolicy")
+	}
+	if len(agentPolicies.Items) == 0 {
```
graphaelli, line 81: Isn't there always a default policy if Elastic Agent is up?  I think this should keep retrying if the default policy is not found.

graphaelli (changes requested):
I really like the installer written in go, particularly for the error handling.  I'm having trouble getting this stack to come up cleanly, consistently.  I believe there is a race between the apm-server-managed container looking for default policy that the elastic-agent container is creating.  A simple depends_on doesn't exactly fix it since a really slow starting elastic-agent will still pass this trivial health check before apm-server-managed tries to find the policy.  I've proposed a retry in the installer to address this.

docker/apm-server/managed/main.go
graphaelli: also, what happens on container restart?  is there a way to check if we're already attached the default policy and do nothing in that case?

docker/apm-server/managed/main.go
```diff
@@ -0,0 +1,245 @@
...
+	http.Handle("/", httputil.NewSingleHostReverseProxy(target))
+	if err := http.ListenAndServe(":8200", nil); err != nil {
+		log.Fatal(err)
+		os.Exit(-1)
+	}
+}
+
+func setup() {
+	client := newKibanaClient()
+	if err := cleanup(client); err != nil {
```
graphaelli, line 32: What if this happens on container restart?  is there a way to check if we're already attached the default policy and do nothing in that case?

docker/apm-server/managed/main.go
```diff
@@ -0,0 +1,245 @@
...
+
+func main() {
+	setup()
+	// create a reverse proxy to the APM Server running via Elastic Agent,
+	// headers are not rewritten, as not considered important right now.
+	target, _ := url.Parse("http://elastic-agent:8200")
+	http.Handle("/", httputil.NewSingleHostReverseProxy(target))
+	if err := http.ListenAndServe(":8200", nil); err != nil {
+		log.Fatal(err)
+		os.Exit(-1)
```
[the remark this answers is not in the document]
simitt: right, 🤦‍♀️

docker/apm-server/managed/main.go
```diff
@@ -0,0 +1,245 @@
...
+}
+
+func newKibanaClient() *kibanaClient {
+	host := os.Getenv("KIBANA_HOST")
+	if host == "" {
+		host = "http://admin:changeme@localhost:5601"
+	}
+	return &kibanaClient{
+		fleetURL: fmt.Sprintf("%s/api/fleet", host),
+		pkgURL:   "https://epr-snapshot.elastic.co",
```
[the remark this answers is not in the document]
jalvz: It should also account for the production (and maybe staging) registry, eg. when running a release branch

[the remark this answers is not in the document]
simitt: good point, so in which cases should we use the release URL - only for `apm-server-release` or als for `apm-server-bc`? (I believe only for release).

docker/apm-server/managed/main.go
simitt: Made some changes and now check if the existing policy fits the expectations - makes the code slightly more complicated, but has a bit faster startup time.

docker/apm-server/managed/main.go
simitt: added some retry logic - I can't seem to slow down the agent enough to run into this; please let me know if you still have issues.

docker/apm-server/managed/main.go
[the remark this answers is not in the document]
simitt: went with `release`  only for now

graphaelli (approved):

kuisathaverat:
@graphaelli @simitt Do we want to enable APM Server managed on the daily ITs?

simitt:
@kuisathaverat the APM Server managed only supports a handful of config options so far. Are there any configurations needed for the daily ITs (except for `apm-server.host` and `apm-server.secret_token`)?

kuisathaverat:
I do not see any other parameter, we can do a try on a PR and see

simitt:
That would be great, thanks @kuisathaverat

PR #1022, opened 2021-01-07
chore: update tested versions

kuisathaverat:
## What does this PR do?

Update versions tested with the APM Server master branch.

### DIFF
--- a/tests/versions/java.yml
+++ b/tests/versions/java.yml
@@ -1,6 +1,4 @@
 JAVA_AGENT:
   - 'github;master'
-  - 'release;1.0.1'
-  - 'release;1.1.0'
-  - 'release;1.2.0'
-  - 'release;1.3.0'
+  - 'release;1.19.0'
+  - 'release;1.18.1'
--- a/tests/versions/python.yml
+++ b/tests/versions/python.yml
@@ -4,5 +4,4 @@ PYTHON_AGENT:
   # release;release -> pip install elastic-apm==version
   - 'github;master'
   - 'release;latest'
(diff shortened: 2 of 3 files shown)


tests/versions/ruby.yml
```diff
@@ -3,5 +3,5 @@ RUBY_AGENT:
   # release;latest -> gem elastic-apm
   # release;version -> gem elastic-apm, version
   - 'github;master'
-  - 'github;2.x'
```
kuisathaverat, line 6: NIT: only changes the order of the versions to follow the same rules for all APM Agents.

tests/versions/python.yml
```diff
@@ -4,5 +4,4 @@ PYTHON_AGENT:
   # release;release -> pip install elastic-apm==version
   - 'github;master'
   - 'release;latest'
-  - 'release;5.3.2'
```
kuisathaverat, line 7: `Latest` should test the 5.x versions

tests/versions/java.yml
```diff
@@ -1,6 +1,4 @@
 JAVA_AGENT:
   - 'github;master'
-  - 'release;1.0.1'
```
kuisathaverat, line 3: really old versions, we add only the latest release

cachedout (approved):

tests/versions/python.yml
basepi: 6.x is upcoming so we'll need to remember to add the last 5.x version here at that time.

basepi (approved):

PR #1027, opened 2021-01-11
[backport] dotnet for 6.x

v1v:
## What does this PR do?

Enable dotnet for the 6.x branch

For such I backported all the changes in 7.x related to the dotnet agent, and fixed some test failures that were related to some missing backports.


v1v:
Test failures can be found in -> https://apm-ci.elastic.co/blue/organizations/jenkins/apm-integration-test-downstream/detail/6.x/3108/tests

gregkalapos:
Regarding the failing test.

Here is what CI says:

```
                    for attr in ['abs_path', 'line', 'filename']:
                        assert all(
>                           frame.get(attr) for frame in stacktrace), stacktrace[0].keys()
E                       AssertionError: dict_keys(['exclude_from_grouping', 'filename', 'abs_path', 'line', 'module', 'function'])
```

This is coming from [here](https://github.com/v1v/apm-integration-testing/blob/feature/support-dotnet-6.x/tests/agent/concurrent_requests.py#L283).

So it seems the .NET Agent is not sending one of the 3 fields the test asserts: `'abs_path', 'line', 'filename'`

From these 3 fields, only `filename` is required, the two other fields are optional. I looked into the agent code, we try to fill the two other fields, but there are cases when we don't get those two - it's just the nature of .NET, if you don't have pdb files or any debug info, we may not be able to get the line number. It would be nice to know what data we get there (e.g. what is in `filename`). I suspect this will be a stack frame from one of the system dlls and in those cases we typically won't get line number info.

Question is, what to do to make this green :) 

There was [this PR](https://github.com/elastic/apm-integration-testing/commit/9e8ffb0ac3bc5bda1da196a54d5670608a810caf#diff-7d088ac0f948d1b5befa1cf36c74214318bc47b275b67b29749bd6f76a5b122e) which removes asserting on optional fields. I think that'd solve the problem here as well. The thing is that I guess those asserts are useful for other agents, so if we remove it, then we remove it for all agents.

So either we accept that we can't assert on optional field in code that is shared across agents (which as it seems is already the case on `master`) or we add some agent check and for .NET.

gregkalapos (approved):

PR #1033, opened 2021-01-20
fix: lint check is broken

kuisathaverat:
## What does this PR do?

Fixes a couple of lint errors.

## Why is it important?

Lint checks are failing

Screenshot 2021-01-20 at 10 03 29

## Related issues
Closes #ISSUE


v1v (approved):
🥇 

There is a corner case:

* Linting is executed for the changeset.
* If it's not fixed and there are more commits in the PR, then the linting won't be detected.

IIRC, we used that approach to avoid linting everything by default, since there might be some defects already there, but only linting the affected file in the changeset.

cachedout (approved):

PR #1037, opened 2021-01-20
OSS distribution change for 7.11+ and 6.8.14+

v1v:
## What does this PR do?

If `--oss` then let's fallback to the `non-oss` distribution for `Kibana` and `Elasticsearch` when `7.11+` and `6.8.14+`

## Why is it important?

As a consequence of the license change, Kibana and Elasticsearch OSS distributions from 7.11 and 6.8.14 won't be available. 

Even though I don't like much to change behaviours under the hood, I've found it's the best way to support the Elasticstack for Kibana, Elasticsearch and APM Server when using the `--oss` flag.

Otherwise we need to change the consumers logic such as:

- https://github.com/elastic/apm-integration-testing/blob/master/tests/versions/apm_server.yml#L3

That particular file is used in our nightly testing with all the APM agents/servers versions.

### DIFF
--- a/scripts/modules/service.py
+++ b/scripts/modules/service.py
@@ -57,6 +57,10 @@ class Service(object):
                 self.apm_api_key = {"ELASTIC_APM_API_KEY": options.get("elastic_apm_api_key")}
             else:
                 print('WARNING: elastic_apm_api_key is not supported for the current version. Use version +7.6.')
+        if self.name() in ("elasticsearch", "kibana"):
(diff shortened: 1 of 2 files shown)


mdelapenya (approved):

cachedout:
I'm not sure about this change. It makes me really nervous when flags end up doing the opposite of what they're supposed to do, even when warnings are supplied.

How much work do you think it would be to change the consumer end of this, like you mention? I think we should further evaluate that option before going down this road.

PR #1040, opened 2021-01-21
OSS distribution change for Kibana and Elasticsearch in 7.11+ and 6.8.14+ versions

v1v:
## What does this PR do?

Fail if unsupported versions for Kibana and Elasticsearch when using the `oss` distribution.

Configure consumer to use only the `apm-server` oss distribution

## Why is it important?

As a consequence of the license change, Kibana and Elasticsearch OSS distributions from 7.11 and 6.8.14 won't be available.

## Related issues

Supersedes https://github.com/elastic/apm-integration-testing/pull/1037

### DIFF
--- a/scripts/modules/service.py
+++ b/scripts/modules/service.py
@@ -57,6 +57,10 @@ class Service(object):
                 self.apm_api_key = {"ELASTIC_APM_API_KEY": options.get("elastic_apm_api_key")}
             else:
                 print('WARNING: elastic_apm_api_key is not supported for the current version. Use version +7.6.')
+        if self.name() in ("elasticsearch", "kibana"):
+            if self.at_least_version("7.11") or (self.at_least_version("6.8.14") and self.version_lower_than("6.9")) :
+                print('WARNING: OSS distribution is not supported in 7.11+/6.8.14+. Unset the oss flag.')
+                self._oss = False
 
     @property
     def bc(self):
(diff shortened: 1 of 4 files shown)


scripts/modules/service.py
```diff
@@ -57,6 +58,10 @@ def __init__(self, **options):
                 self.apm_api_key = {"ELASTIC_APM_API_KEY": options.get("elastic_apm_api_key")}
             else:
                 print('WARNING: elastic_apm_api_key is not supported for the current version. Use version +7.6.')
+        if self._oss and self.name() in ("elasticsearch", "kibana"):
+            if self.at_least_version("7.11") or (self.at_least_version("6.8.14") and self.version_lower_than("6.9")):
+                print('ERROR: OSS distribution is ONLY supported in 7.11+/6.8.14+ for Kibana and Elasticsearch.')
+                sys.exit(1)
```
cachedout, line 64: 💯 Thanks for acknowledging my paranoia. ;) This makes me feel much better.

cachedout (approved):

scripts/modules/service.py
v1v: I was not 100% confident about my first attempt, thanks for challenging it!!

tests/versions/apm_server.yml
```diff
@@ -1,4 +1,4 @@
 APM_SERVER:
   - master
-  - master --oss
+  - master --apm-server-oss
```
kuisathaverat, line 3: not sure if APM Server will have only one binary too

v1v: IIUC, beats didn't transition to a different license but Elasticsearch and Kibana, so I'd assume no changes in that particular regard

kuisathaverat (approved):

PR #1069, opened 2021-02-24
Add --package-registry-url flag

### LINKED ISSUE #1068: Cannot run 7.x build candidates with snapshot registry
I'd like to test the 7.12 build candidates with Fleet, using the APM package in the snapshot package registry. When running a build candidate of Kibana, it will interact with the production package registry by default. This can be overridden with `XPACK_FLEET_REGISTRYURL` (`xpack.fleet.registryUrl`).

When running the managed APM Server service, one can pass the `--package-registry-snapshot` flag to direct it to use the snapshot registry. However, this only sets `XPACK_FLEET_REGISTRYURL` for the managed APM Server service. It should also be set for the Kibana process, so Kibana can fetch packages from the appropriate registry.

axw:
## What does this PR do?

Add a `--package-registry-url` flag to control the Elastic Package Registry URL passed into the managed APM Server service and Kibana service environment variables.

Also, set `XPACK_FLEET_REGISTRYURL` in the Kibana environment (consistently with APM Server) regardless of whether we're using a local package registry or not.

Remove the dependency from APM Server to the local package registry; there is already a transitive dependency through Kibana.

## Why is it important?

When running build candidates in the future we should probably default to running with the production Elastic Package Registry (epr.elastic.co). We will need to test with snapshots and staging though, so we can test with packages before they're promoted.

## Related issues

Closes #1068

### DIFF
--- a/scripts/modules/elastic_stack.py
+++ b/scripts/modules/elastic_stack.py
@@ -42,14 +42,8 @@ class ApmServer(StackService, Service):
                 kibana_url = kibana_scheme + "://admin:changeme@" + self.DEFAULT_KIBANA_HOST
             self.depends_on = {"kibana": {"condition": "service_healthy"}}
 
-            xpack_registry_url = "https://epr-snapshot.elastic.co"
-            if options.get("enable_package_registry"):
-                xpack_registry_url = "http://package-registry:{}".format(PackageRegistry.SERVICE_PORT)
-                self.depends_on["package-registry"] = {"condition": "service_healthy"}
-            elif options.get("release") or options.get("apm_server_release"):
-                xpack_registry_url = "https://epr.elastic.co"
             self.managed_environment = {"KIBANA_HOST": kibana_url,
-                                        "XPACK_FLEET_REGISTRYURL": xpack_registry_url,
+                                        "XPACK_FLEET_REGISTRYURL": package_registry_url(options),
                                         "APM_SERVER_SECRET_TOKEN": self.options.get("apm_server_secret_token", "")}
 
             # Not yet supported when run under elastic-agent:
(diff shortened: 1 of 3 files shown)


kuisathaverat (approved):

jalvz:
this is very similar to https://github.com/elastic/apm-integration-testing/blob/master/scripts/modules/elastic_stack.py#L678-L682, isn't it?

axw:
@jalvz it is very similar, but that one just controls the local package registry. It could be useful to run a proxy between Kibana and EPR, so I think we should allow a full URL to be specified. Maybe in another PR we could change the `--package-registry-distribution` flag to control the default value for EPR as well as the local distribution.

scripts/modules/elastic_stack.py
```diff
@@ -42,14 +42,8 @@ def __init__(self, **options):
                 kibana_url = kibana_scheme + "://admin:changeme@" + self.DEFAULT_KIBANA_HOST
             self.depends_on = {"kibana": {"condition": "service_healthy"}}
 
-            xpack_registry_url = "https://epr-snapshot.elastic.co"
-            if options.get("enable_package_registry"):
-                xpack_registry_url = "http://package-registry:{}".format(PackageRegistry.SERVICE_PORT)
-                self.depends_on["package-registry"] = {"condition": "service_healthy"}
```
simitt, line 48: Did you remove the healthcheck on purpose?

simitt (commented):
Shouldn't the `enable_package_registry` be aware of the `package_registry_url` ?

scripts/modules/elastic_stack.py
axw: Yes:

> Remove the dependency from APM Server to the local package registry; there is already a transitive dependency through Kibana.

axw:
> Shouldn't the enable_package_registry be aware of the package_registry_url ?

Not sure I follow. The new `package_registry_url` function consults `enable_package_registry`

simitt (approved):
All good, I got confused by multiple package registry related config options.

PR #1078, opened 2021-03-09
Fix APM Server managed by Elastic Agent setup

simitt:
The latest docker builds for Kibana and Elastic Agent for `master` and `7.13` are working again; the apm-integration testing setup needs to be adjusted to the latest fixes. 

There seems to be something up with the production package registry - it is not serving an APM package. Snapshot registry works fine. 

This PR includes:
* print statements for APM Server setup for easier debugging in case something goes wrong
* APM Server setup uses Fleet API now for fetching APM package instead of directly calling the EPR. The configured EPR URL already needs to be considered by Kibana. 
* Adding APM Server integration to Fleet Policy, to which the Elastic Agent now enrolls. 
* Using Snapshot EPR if `--snapshot` flag is provided, rather than production one.

### DIFF
--- a/docker/apm-server/managed/main.go
+++ b/docker/apm-server/managed/main.go
@@ -37,6 +37,7 @@ func setupManagedAPM() error {
 	if err != nil {
 		return err
 	}
+	fmt.Println("default policy fetched")
 
 	// fetch the available apm package
 	packages, err := client.getPackages("package=apm&experimental=true")
@@ -46,6 +47,7 @@ func setupManagedAPM() error {
 	if len(packages) == 0 {
 		return errors.New("no apm package found")
 	}
+	fmt.Println("apm package fetched")
 
 	// define expected APM package policy
 	expectedAPMPackagePolicy := apmPackagePolicy(policy.ID, packages[0])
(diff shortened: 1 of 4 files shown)


stuartnelson3 (approved):
looks fine to me

scripts/modules/elastic_stack.py
```diff
@@ -1185,7 +1186,7 @@ def package_registry_url(options):
         return url
     if options.get("enable_package_registry"):
         return "http://package-registry:{}".format(PackageRegistry.SERVICE_PORT)
-    elif options.get("snapshot"):
+    elif options.get("snapshot") or not options.get("release"):
```
simitt, line 1189: @axw this changes the default to snapshot, which will serve snapshot+staging+production packages; just wanted to get your 👀 on this real quick.

axw: I suppose it's fine. My thinking was that we would rely on the logic in Kibana to select the appropriate registry, though to be fair we were still diverging by selecting the snapshot registry on snapshot stack images. But we can always select the staging repo if needed, so :+1:

docker/apm-server/managed/main.go
```diff
@@ -155,31 +163,33 @@ func apmPackagePolicy(policyID string, pkg eprPackage) packagePolicy {
...
+			continue
+		}
+		var apm eprPackageResponse
+		url := fmt.Sprintf("%s/epm/packages/%s-%s", client.fleetURL, p.Name, p.Version)
+		err := makeRequest(http.MethodGet, url, nil, &apm)
+		return &apm.Package, err
+
 	}
-	return packages, nil
+	return nil, nil
```
axw, line 192: nit:
 
```suggestion
	return nil, errors.New("no apm package found")
```

then you can get rid of the nil check in the caller

axw (approved):

PR #1082, opened 2021-03-17
Add terraform templates for deploying Azure resources

russcam:
## What does this PR do?

This PR adds a terraform template for deploying Azure Service
Bus resources for integration testing purposes. Included is a README
providing a quickstart guide to get up and running.

## Why is it important?

To be able to run integration tests against real Azure resources.


kuisathaverat:
I am not sure if we should put this terraform stuff here, it is not related to compose.py, @elastic/observablt-robots WDYT?

russcam:
They're not related to compose.py but are very much related to APM integration testing, for use by APM agents to test against real cloud provider resources. It's likely that the number of terraform templates will grow as agents implement integrations against AWS, Azure, GCP, etc.

cachedout:
> I am not sure if we should put this terraform stuff here

From reading the description, it sounds like this is a PR which lays the groundwork for future PRs which will specifically add automated integrations tests which make use of the Terraform templates it includes. @russcam  Is that correct?

Assuming so, I can see the value of including the templates in this project, but I am not certain that they should exist at the top-level of the directory hierarchy. If they are exclusively for the use of the test suite, perhaps they would be better situated under the `tests/` directory?

russcam:
> From reading the description, it sounds like this is a PR which lays the groundwork for future PRs which will specifically add automated integrations tests which make use of the Terraform templates it includes. @russcam Is that correct?

In future, it could form part of automated integration tests within this repository but for now, this is a central location for sharing terraform templates amongst APM agents for integration testing running both locally and in their respective CI environments.

kuisathaverat:
the thing is that this repo host the integration between the APM Server and the Agents, it does not host any other integration test types. This terraform definition is an integration between the .Net Agent and Azure.

russcam:
> This terraform definition is an integration between the .Net Agent and Azure.

The .NET agent is the first to use them, but they are for **any** agent to use, which they'll need when it comes to testing their Azure integration implementations. This repository seems like the obvious place to put shared resources for APM integration testing, but if there's a strong consensus that this isn't the right place for them, they can go into a new repository.

adam-stokes:
> The .NET agent is the first to use them, but they are for **any** agent to use, which they'll need when it comes to testing their Azure integration implementations. This repository seems like the obvious place to put shared resources for APM integration testing, but if there's a strong consensus that this isn't the right place for them, they can go into a new repository.

This is just a general question, could this be useful for other projects wanting to test integrations? We have https://github.com/elastic/e2e-testing that currently utilizes docker for setup/teardown of environments to test elastic-agent, fleet, metricbeat, etc. But, it would be cool to expand that scope to cloud substrates as well, could the terraform templates be expanded to allow for that sort of thing as well? If so, would it make more sense to have a terraform deployment repo designed in a way that could support those different type of configurations?

russcam:
The terraform templates can be generally used as-is for testing against cloud providers, so they could be useful for other projects. We can start another repository with templates if you'd also like to use and contribute to them.

kuisathaverat:
I think it fits on the obtl test environments repo

russcam:
> I think it fits on the obtl test environments repo

This one? https://github.com/elastic/observability-test-environments

kuisathaverat:
yes, that repo. The terraform templates in the PR are only part of the requirements to deploy those resources on Azure, you need an account and download some tools. On the bolt teste environments repo, we have a wrapper on top of deployments making them easier for developers, you checkout the repo, configure a simple YAML file and run a command, the account used, the tools, the error manage, and so on is out of the box. We are working on integrate terraform deploys on the list of support deployments as soon as this is done I will copy these templates there, and document how to use these an others terraform templates we will have.

* [Azure Cosmos DB](https://docs.microsoft.com/en-us/azure/cosmos-db/introduction)
* [Azure Service Bus](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-messaging-overview)
* [Azure Storage services](https://docs.microsoft.com/en-us/azure/storage/common/storage-introduction)

adam-stokes:
Closing in favor of https://github.com/elastic/observability-test-environments/pull/1651

Issue #1083, opened 2021-03-22
"--opbeans-node-agent-local-repo=/some/local/dir" doesn't work


trentm:
The compose.py script supports some `--opbeans-$lang-agent-local-repo=$localDir` options to have the opbeans-$lang container use an APM agent from the given `$localDir` (a working copy of the APM agent repo). For example:

```bash
% ./scripts/compose.py start master --with-opbeans-node \
    --opbeans-node-agent-local-repo=~/el/apm-agent-nodejs5
```

Two issues using the node.js version of this option:

1. It doesn't actually install the Node.js APM agent from the local repo in a way that the opbeans-node installed at "/app" uses it.
2. Part of the handling for this involves making a copy of the local repo dir (to not pollute it). Unfortunately the local repo dir can have a somewhat weighty node_modules dir, which can take a long time (sometimes too long, resulting in the container being marked as unhealthy).

![Screen Shot 2021-03-22 at 1 39 58 PM](https://user-images.githubusercontent.com/46866/112055541-39c8cd80-8b14-11eb-8abc-be32af160b90.png)

In my case:

```
% docker exec -ti 699d48159443 /bin/sh
/app # cd ~/local-install/
~/local-install # du -sh .
484.3M	.
~/local-install # du -sh node_modules
370.9M	node_modules
~/local-install # du -sk * | sort -n | tail -3
1516	test
101288	build
379756	node_modules
```

That's copying almost half a GiB for every `./scripts/compose.py start ...`.

PR coming to attempt to fix these.

PR #1084, opened 2021-03-22
fix: --opbeans-node-agent-local-repo=/a/local/dir

trentm:
## What does this PR do?

Fixes './scripts/compose.py start --opbeans-node-agent-local-repo=$dir'
to (a) install the Node.js APM agent into the opbeans-node app at "/app"
and (b) to skip copying the possibly large node_modules and build dirs
for speed.

## Why is it important?

Otherwise the starting of the opbeans-node container is potentially very slow and doesn't run the Node.js APM agent code one expects.

## Related issues

Fixes: #1083


docker/opbeans/node/entrypoint.sh
```diff
@@ -1,11 +1,14 @@
 #!/usr/bin/env sh
 set -ex
 if [ -f /local-install/package.json ]; then
-    echo "Installing from local folder"
-    # copy to folder inside container to ensure were not poluting the local folder
-    cp -r /local-install ~
-    cd ~/local-install && npm install .
```
trentm, line 7: This `npm install .` was installing the deps for the apm-agent-nodejs copy in ~/local-install, but it wasn't installing them where the app at "/app" would use it.

docker/opbeans/node/entrypoint.sh
```diff
@@ -1,11 +1,14 @@
...
-    cd -
+    echo "Installing elastic-apm-node from local folder (--opbeans-node-agent-local-repo)"
+    # Copy to a folder inside container to ensure we're not polluting the
+    # local folder. Skip possibly huge dirs to speed this up.
+    apk add rsync
```
trentm, line 7: Is installing rsync appropriate here?  Would doing so in the Dockerfile be more appropriate?

kuisathaverat: it is better to add packages on the Dockerfile

v1v (approved):

kuisathaverat (approved):

PR #1085, opened 2021-03-23
fix: update flake8 and deps to work with Python 3.8+

trentm:
This fixes 'make lint' when one has Python 3.8 or later
(https://github.com/PyCQA/pyflakes/issues/367), but updating to the
latest flake8 and its deps. Note that flake8 3.9.0 dropped support
for Python 3.4
(https://flake8.pycqa.org/en/latest/release-notes/3.9.0.html)

In my case I have Python 3.9. Presumably CI is using a Python version less than 3.8?
Is support for Python 3.4 required? If so, then we can't upgrade the flake8 v3.9.0.

Before:

```
% make lint
test -d ./venv || virtualenv -q --python=python3 ./venv;\
	source ./venv/bin/activate || exit 1;\
	pip install -q -r requirements.txt;\
	touch ./venv;\

flake8 --ignore=D100,D101,D102,D103,D104,D105,D106,D107,D200,D205,D400,D401,D403,W504  tests/ scripts/compose.py scripts/modules
multiprocessing.pool.RemoteTraceback:
"""
Traceback (most recent call last):
  File "/Users/trentm/tm/apm-integration-testing/venv/lib/python3.9/site-packages/flake8/checker.py", line 451, in run_check
    return plugin["plugin"](**arguments)
  File "/Users/trentm/tm/apm-integration-testing/venv/lib/python3.9/site-packages/flake8/plugins/pyflakes.py", line 99, in __init__
    super(FlakesChecker, self).__init__(
  File "/Users/trentm/tm/apm-integration-testing/venv/lib/python3.9/site-packages/pyflakes/checker.py", line 497, in __init__
    self.handleChildren(tree)
  File "/Users/trentm/tm/apm-integration-testing/venv/lib/python3.9/site-packages/pyflakes/checker.py", line 824, in handleChildren
    self.handleNode(node, tree)
  File "/Users/trentm/tm/apm-integration-testing/venv/lib/python3.9/site-packages/pyflakes/checker.py", line 871, in handleNode
    handler(node)
  File "/Users/trentm/tm/apm-integration-testing/venv/lib/python3.9/site-packages/pyflakes/checker.py", line 824, in handleChildren
    self.handleNode(node, tree)
  File "/Users/trentm/tm/apm-integration-testing/venv/lib/python3.9/site-packages/pyflakes/checker.py", line 871, in handleNode
    handler(node)
[...]

PR #1086, opened 2021-03-24
fix: force to activate Python virtualenv

kuisathaverat:
## What does this PR do?

* It forces to activate the Python virtualenv 

## Why is it important?

The trick to continue the one-line command for a previous target does not work in some cases, so the virtual environment is not used.

## Related issues
Related to https://github.com/elastic/apm-integration-testing/pull/1085#issuecomment-805616752

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,9 +54,10 @@ venv: requirements.txt ## Prepare the virtual environment
 	test -d $(VENV) || virtualenv -q --python=$(PYTHON) $(VENV);\
 	source $(VENV)/bin/activate || exit 1;\
 	pip install -q -r requirements.txt;\
-	touch $(VENV);\
+	touch $(VENV);
 
 lint: venv  ## Lint the project
+	source $(VENV)/bin/activate; \
 	flake8 --ignore=D100,D101,D102,D103,D104,D105,D106,D107,D200,D205,D400,D401,D403,W504  tests/ scripts/compose.py scripts/modules
 
 .PHONY: create-x509-cert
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -54,9 +54,10 @@ venv: requirements.txt ## Prepare the virtual environment
 	test -d $(VENV) || virtualenv -q --python=$(PYTHON) $(VENV);\
 	source $(VENV)/bin/activate || exit 1;\
 	pip install -q -r requirements.txt;\
-	touch $(VENV);\
+	touch $(VENV);
 
 lint: venv  ## Lint the project
+	source $(VENV)/bin/activate; \
```
v1v, line 60: L55 does the activate -> https://github.com/elastic/apm-integration-testing/pull/1086/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R55

Do you think the `venv` goal could be split in two, then the consumers don't need to do anything else?


```
prepare-venv: requirements.txt ## Prepare the virtual environment
	test -d $(VENV) || virtualenv -q --python=$(PYTHON) $(VENV);

venv: prepare-venv  ## Prepare the virtual environment
	source $(VENV)/bin/activate || exit 1;\
	pip install -q -r requirements.txt;\
	touch $(VENV);
```

kuisathaverat: I do not see advantages

v1v: You don't need to change the consumers, so the other goals remain as they are

kuisathaverat: I just tested it, have the same issue, it does not activate the environment correctly in some cases. I have tried these two codes:

```
prepare-venv: requirements.txt ## Prepare the virtual environment
	test -d $(VENV) || virtualenv -q --python=$(PYTHON) $(VENV);

venv: prepare-venv  ## Prepare the virtual environment
	source $(VENV)/bin/activate || exit 1;\
	pip install -q -r requirements.txt;\
	touch $(VENV);
```

```
prepare-venv: requirements.txt ## Prepare the virtual environment
	test -d $(VENV) || virtualenv -q --python=$(PYTHON) $(VENV);

venv: prepare-venv  ## Prepare the virtual environment
	source $(VENV)/bin/activate || exit 1;\
	pip install -q -r requirements.txt;\
	touch $(VENV);\
```

I think is better to be explicit and avoid magic, when magic stop working you do not realize what's going on.

v1v (approved):

PR #1104, opened 2021-04-07
fix: drop ancient and conflicting maxQueueSize setting for node.js agent

trentm:
The maxQueueSize and flushInterval vars are settings from v1 of the
node.js agent -- EOL'd more than 2 years ago:
https://www.elastic.co/guide/en/apm/agent/nodejs/current/upgrade-to-v2.html#v2-removed-config-options

Now, https://github.com/elastic/apm-agent-nodejs/pull/2024 is adding a
maxQueueSize config variable with a different meaning. This old
`maxQueueSize: 1` breaks integration tests for the new code by telling
the agent to drop any transactions/span/errors if there is already a
single event that hasn't yet been sent to the APM server.

### DIFF
--- a/docker/nodejs/express/app.js
+++ b/docker/nodejs/express/app.js
@@ -4,8 +4,6 @@ var apm = require('elastic-apm-node').start({
   frameworkName: 'express',
   frameworkVersion: 'unknown',
   serviceName: process.env.EXPRESS_SERVICE_NAME,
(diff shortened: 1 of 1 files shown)


astorm (approved):
The `maxQueueSize` is dead, long live the`maxQueueSize`. 

Removing these from the services used by the integration tests makes sense.  Approved.

PR #1105, opened 2021-04-07
fix: correct Makefile syntax to spawn a subshell

trentm:
This fixes a buglet (IIUC) in the Makefile `destroy-env` target.  With the single `$` it is trying to do variable expansion, which resolves tothe empty string and you get a no-op everytime:

```
% make destroy-env
[ -n "" ] && (docker ps -aqf network=apm-integration-testing | xargs -t docker rm -f && docker network rm apm-integration-testing) || true
```

(Sorry for the branch on the main repo, I hit the wrong option with `gh pr create ...` that pushed to this repo instead of to my fork.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,7 +81,7 @@ stop-env: venv ## Stop the test environment
 	docker-compose down -v --remove-orphans || true
 
 destroy-env: venv ## Destroy the test environment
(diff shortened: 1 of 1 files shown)


trentm:
This syntax was originally added in #271

kuisathaverat (approved):

PR #1115, opened 2021-04-28
fix: grab stack version from the library

kuisathaverat:
## What does this PR do?

this changes the param `ELASTIC_STACK_VERSION` to use the stack version defined in the library by default.

## Why is it important?

The step `stackVersions` is updated on FF and Releases so it contains the latest version we are developing and releasing.


cachedout (approved):

axw (approved):

mdelapenya (approved):

kuisathaverat:
errors not related to the change

PR #1116, opened 2021-04-28
fix: grab stack version from the library

kuisathaverat:
## What does this PR do?

this changes the param `ELASTIC_STACK_VERSION` to use the stack version defined in the library by default.

## Why is it important?

The step `stackVersions` is updated on FF and Releases so it contains the latest version we are developing and releasing.

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -16,6 +16,7 @@ pipeline {
     JOB_GCS_BUCKET = credentials('gcs-bucket')
     PIPELINE_LOG_LEVEL='INFO'
     GIT_REFERENCE_REPO = '/var/lib/jenkins/.git-references/apm-integration-testing.git'
+    ELASTIC_STACK_VERSION = "${ params?.ELASTIC_STACK_VERSION?.trim() ? params.ELASTIC_STACK_VERSION.trim() : stackVersions.edge() }"
   }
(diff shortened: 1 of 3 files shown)


cachedout (approved):

.ci/Jenkinsfile
```diff
@@ -16,6 +16,7 @@ pipeline {
     JOB_GCS_BUCKET = credentials('gcs-bucket')
     PIPELINE_LOG_LEVEL='INFO'
     GIT_REFERENCE_REPO = '/var/lib/jenkins/.git-references/apm-integration-testing.git'
+    ELASTIC_STACK_VERSION = "${ params?.ELASTIC_STACK_VERSION?.trim() ? params.ELASTIC_STACK_VERSION.trim() : stackVersions.edge() }"
```
v1v, line 19: `stackVersions.edge()`, will it work? IIRC, if it was a step then it requires the `script` closure, just in case

kuisathaverat: Interpolations are scripts, this pipeline works as expected

```
pipeline {
    agent { label 'master'}
    environment {
        FOO = "${stackVersions.release()}"
    }
    stages {
        stage('Hello') {
            steps {
                echo "Hello World ${FOO}"
            }
        }
    }
}
```

PR #1136, opened 2021-05-05
fix(README): add linebreak to bash command

thomasklinger1234:
## What does this PR do?

Fixes a syntax error when starting a local dev environment. The missing part here is the `\` linebreak that is now added.

## Why is it important?

Copy & paste does not work.


cachedout (approved):

cachedout:
Nice catch. Thanks, @thomasklinger1234 !

kuisathaverat (approved):

PR #1147, opened 2021-05-12
fix --no-apm-server-profile

beniwohli:
## What does this PR do?

The wrong flag name was used when checking its value. This fixes it.

## Why is it important?

Agent devs might want to disable apm-server self-profiling to not have the profiling index filled with distracting data.

### DIFF
--- a/scripts/modules/elastic_stack.py
+++ b/scripts/modules/elastic_stack.py
@@ -99,7 +99,7 @@ class ApmServer(StackService, Service):
 
         if options.get("apm_server_self_instrument", True):
             self.apm_server_command_args.append(("apm-server.instrumentation.enabled", "true"))
-            if self.at_least_version("7.6") and options.get("no_apm_server_profile", False) == False:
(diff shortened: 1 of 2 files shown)


kuisathaverat (approved):

scripts/modules/elastic_stack.py
```diff
@@ -90,7 +90,7 @@ def __init__(self, **options):
         ])
         if options.get("apm_server_self_instrument", True):
             self.apm_server_command_args.append(("apm-server.instrumentation.enabled", "true"))
-            if self.at_least_version("7.6") and options.get("apm_server_profile", True):
+            if self.at_least_version("7.6") and options.get("no_apm_server_profile", False):
```
kuisathaverat, line 93: double negation option, I have had to read it twice 😄

v1v:
@beniwohli , there are broken UTs with this change.

I also labelled to backport this change to the 7.x branch, please feel free to amend it if that's not the expected requirement

PR #1208, opened 2021-07-13
uninstall any previously installed version of elastic-apm in entrypoint.sh

beniwohli:
## What does this PR do?
pip doesn't reinstall a package if it has the same version as the already
installed package. If the requested repo/branch has the same package version,
pip will download it, but not install it.

To avoid this issue altogether, we uninstall the version installed through
the Dockerfile first.

## Why is it important?

It makes the `--opbeans-python-agent-branch` and `--opbeans-python-agent-repo` flags work as expected in every case.


basepi (approved):
Excellent! I feel like I've been burned enough times that I haven't been using this as much as I should, so this will be a good fix.

PR #1209, opened 2021-07-16
Fix dotnet concurrency tests

russcam:
## What does this PR do?

This commits fixes the dotnet concurrency tests by disabling metrics and cloud provider metadata discovery, through agent configuration.

The concurrency tests are failing an assertion on the count of expected spans, where the failure is typically a difference of one or two spans. What I believe is happening is that the volume of metrics produced within tests is flooding the payload queue such that when the application shuts down, there are spans remaining in the queue that do not end up getting sent to APM server.

The agent does not currently drain remaining items from the payload queue to send on shutdown, which is something to consider implementing in future to help with this issue.

## Why is it important?

The intermittent failing dotnet concurrency tests are causing unrelated PRs to fail

## Related issues
Closes https://github.com/elastic/apm-agent-dotnet/issues/1353

### DIFF
--- a/docker/dotnet/aspnetcore/Program.cs
+++ b/docker/dotnet/aspnetcore/Program.cs
@@ -1,5 +1,6 @@
 ﻿using Microsoft.AspNetCore;
 using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.Logging;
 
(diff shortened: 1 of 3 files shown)


cachedout:
@Mergifyio backport 6.x

PR #1220, opened 2021-07-22
fix: remove EOL versions

kuisathaverat:
## What does this PR do?

* remove two versions that raise its EOL

image

## Why is it important?

We no need to test them any longer. 

https://www.elastic.co/support/eol

### DIFF
--- a/tests/versions/apm_server.yml
+++ b/tests/versions/apm_server.yml
@@ -8,5 +8,3 @@ APM_SERVER:
   - '7.8;--release'
   - '7.7;--release'
   - '7.6;--release'
-  - '7.5;--release'
-  - '7.4;--release'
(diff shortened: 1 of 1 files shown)


mdelapenya (approved):

cachedout (commented):
I don't think we can merge this. See my explanation here: https://github.com/elastic/observability-robots/discussions/689#discussioncomment-1037096

tests/versions/apm_server.yml
```diff
@@ -8,5 +8,3 @@ APM_SERVER:
   - '7.8;--release'
   - '7.7;--release'
   - '7.6;--release'
```
kuisathaverat, line 10:
```suggestion
```

PR #1222, opened 2021-07-29
feat: use terraform deployment

kuisathaverat:
## What does this PR do?

It changes to use terraform deployments.

## Why is it important?

Terraform ec provider is the new way to provision clusters, it improves the reliability and the speed of the deploys.


mdelapenya (approved):

mdelapenya:
I see a lot of benefits on standardising the deployment with terraform across projects. As discussed yesterday, we are evaluating using it for the e2e-testing repo: deploying the runtime deps (stack + fleet-server) on cloud will simplify the project and deployment options.

Let's discuss if it's possible to wrap it in a step so that we can reuse it 🚀

cachedout (approved):
🥇 Nice!

PR #1229, opened 2021-08-18
add --apm-server-enable-data-streams option

graphaelli:
## What does this PR do?

adds `--apm-server-enable-data-streams` option for instructing APM Server to write to data streams.  This is valid in 7.16 and 8.0 (and maybe 7.15? it's not important to prevent it from being set in earlier versions)

## Related issues
related to https://github.com/elastic/apm-server/issues/5949 - apm-server remains unhealthy until the integration is installed, not ideal but not blocked on that.

### DIFF
--- a/scripts/tests/service_tests.py
+++ b/scripts/tests/service_tests.py
@@ -2,7 +2,6 @@ from __future__ import print_function
 
 import json
 import os
-import pytest
 import unittest
 
 import yaml
--- a/scripts/tests/service_tests.py
+++ b/scripts/tests/service_tests.py
@@ -2,6 +2,7 @@ from __future__ import print_function
 
 import json
 import os
+import pytest
 import unittest
 
 import yaml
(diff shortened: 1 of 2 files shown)


scripts/tests/service_tests.py
```diff
@@ -917,6 +917,10 @@ def test_apm_server_custom_pipeline(self):
         self.assertIn("foo:/usr/share/apm-server/ingest/pipeline/definition.json", apm_server["volumes"])
         self.assertIn("apm-server.register.ingest.pipeline.overwrite=true", apm_server["command"])
 
+    def test_data_streams(self):
+        apm_server = ApmServer(version="7.16.0", apm_server_enable_data_streams=True).render()["apm-server"]
```
simitt, line 921: Why not allowing from `7.15` on?

graphaelli: I've removed any restriction on version number, this is just the one I picked for the test case.

simitt (approved):

Issue #1232, opened 2021-08-26
Kibana 8.0.0 not starting


graphaelli:
Plain old `./scripts/compose.py start 8.0.0` fails to bring up an 8.0.0-SNAPSHOT cluster as of at least 2021-08-24.  Kibana logs:

```
localtesting_8.0.0_kibana | {"type":"log","@timestamp":"2021-08-26T00:52:06+00:00","tags":["info","savedobjects-service"],"pid":1195,"message":"[.kibana] INIT -> CREATE_NEW_TARGET. took: 232ms."}
localtesting_8.0.0_kibana | {"type":"log","@timestamp":"2021-08-26T00:52:06+00:00","tags":["error","savedobjects-service"],"pid":1195,"message":"[.kibana] Action failed with 'security_exception: [security_exception] Reason: action [indices:admin/create] is unauthorized for user [kibana_system_user] with roles [kibana_system] on indices [.kibana_8.0.0_001], this action is granted by the index privileges [create_index,manage,all]'. Retrying attempt 1 in 2 seconds."}
```

The same message is logged for `.kibana_task_manager` and `.kibana_task_manager_8.0.0_001`.

There appears to be some change to permissions around the kibana system indices.  https://github.com/elastic/elasticsearch/pull/76449 looks suspicious but is 2 weeks old and doesn't seem to change this exact behavior.

I attempted to force the permissions with the following to no avail:

```
diff --git b/docker/elasticsearch/roles.yml a/docker/elasticsearch/roles.yml
index 7eb127d..59228f5 100644
--- b/docker/elasticsearch/roles.yml
+++ a/docker/elasticsearch/roles.yml
@@ -23,6 +23,10 @@ heartbeat:
   indices:
     - names: ['heartbeat-*', 'shrink-heartbeat-*']
       privileges: ['all']
+kibana_system_extra:
+  indices:
+    - names: ['.kibana', '.kibana_*', '.kibana_task_manager', '.kibana_task_manager_8.0.0_001']
+      privileges: ['all']
 metricbeat:
   cluster: ['manage_index_templates', 'monitor', 'manage_ingest_pipelines', 'manage_ilm']
   indices:
diff --git b/docker/elasticsearch/users_roles a/docker/elasticsearch/users_roles
index 629fe73..1b6a793 100644
--- b/docker/elasticsearch/users_roles
+++ a/docker/elasticsearch/users_roles
@@ -7,6 +7,7 @@ filebeat:filebeat_user
[...]

PR #1233, opened 2021-08-26
fix apm-server when data_streams enabled

graphaelli:
## What does this PR do?

see https://github.com/elastic/apm-integration-testing/pull/1229#issuecomment-906747371, in summary this changes:

* configure apm-server to use a superuser when communicating to kibana (used for checking integration installation status)
* use the integration pipeline id instead of apm
* allows writes to data stream patterns

## Related issues
This relies on a fix to ES that has not been published in a snapshot yet.  see https://github.com/elastic/apm-integration-testing/issues/1232#issuecomment-906714862

### DIFF
--- a/docker/elasticsearch/roles.yml
+++ b/docker/elasticsearch/roles.yml
@@ -2,7 +2,7 @@
 apm_server:
   cluster: ['manage_ilm', 'manage_security', 'manage_api_key']
   indices:
-    - names: ['apm-*']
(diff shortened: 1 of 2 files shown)


simitt (approved):
Thanks!

graphaelli:
Noting here for future us - we tried to add `XPACK_FLEET_PACKAGES=[{"name":"apm", "version":"latest"}]` to the kibana container's environment when `--apm-server-enable-data-streams` but kibana didn't like it, so the integration package needs to be installed manually for now.

Issue #1188, opened 2021-06-30
Error while running the ALL stage on CI builds


mdelapenya:
It seems it's the apm-server container, which is not healthy on compose startup:

```
[2021-06-30T03:26:23.880Z] ERROR: for apm-server  Container "d609427c2b3f" is unhealthy.
[2021-06-30T03:26:23.880Z] Encountered errors while bringing up the project.
[2021-06-30T03:26:23.880Z] Starting/Building stack services..
[2021-06-30T03:26:23.880Z] 
[2021-06-30T03:26:23.880Z] Traceback (most recent call last):
[2021-06-30T03:26:23.880Z]   File "scripts/compose.py", line 31, in <module>
[2021-06-30T03:26:23.880Z]     main()
[2021-06-30T03:26:23.880Z]   File "scripts/compose.py", line 17, in main
[2021-06-30T03:26:23.880Z]     setup()
[2021-06-30T03:26:23.880Z]   File "/var/lib/jenkins/workspace/tegration-test-downstream_master/src/github.com/elastic/apm-integration-testing/scripts/modules/cli.py", line 204, in __call__
[2021-06-30T03:26:23.880Z]     self.args.func()
[2021-06-30T03:26:23.880Z]   File "/var/lib/jenkins/workspace/tegration-test-downstream_master/src/github.com/elastic/apm-integration-testing/scripts/modules/cli.py", line 586, in start_handler
[2021-06-30T03:26:23.880Z]     self.build_start_handler("start")
[2021-06-30T03:26:23.880Z]   File "/var/lib/jenkins/workspace/tegration-test-downstream_master/src/github.com/elastic/apm-integration-testing/scripts/modules/cli.py", line 732, in build_start_handler
[2021-06-30T03:26:23.880Z]     self.run_docker_compose_process(docker_compose_cmd + up_params)
[2021-06-30T03:26:23.880Z]   File "/var/lib/jenkins/workspace/tegration-test-downstream_master/src/github.com/elastic/apm-integration-testing/scripts/modules/cli.py", line 472, in run_docker_compose_process
[2021-06-30T03:26:23.880Z]     subprocess.check_call(docker_compose_cmd)
[2021-06-30T03:26:23.880Z]   File "/usr/lib/python3.6/subprocess.py", line 311, in check_call
[2021-06-30T03:26:23.880Z]     raise CalledProcessError(retcode, cmd)
[...]

PR #1242, opened 2021-08-31
Start PHP/Python tests with --no-kibana

cachedout:
## What does this PR do?

Resolves failures in the master branch.

No other test suites use Kibana and in inspecting these test suites, neither do Python or PHP. I ran the test suites manually both with and without the the flags in question and the results were the same. Additionally, I looked over the tests themselves and didn't see anything jump out at me suggesting that Kibana would be required to run these tests in CI.

## Why is it important?

Resolves failures in master branch and also slightly reduces the load we put on CI workers. 

## Related issues
Closes https://github.com/elastic/apm-integration-testing/issues/1188

### DIFF
--- a/.ci/scripts/common.sh
+++ b/.ci/scripts/common.sh
@@ -46,6 +46,7 @@ function prepareAndRunAll() {
     --no-xpack-secure \
     --apm-server-enable-tls \
     --no-verify-server-cert  \
+    --no-kibana \
     --apm-server-secret-token=${ELASTIC_APM_SECRET_TOKEN} \
     --apm-server-url=${APM_SERVER_URL} \
     --apm-log-level=debug"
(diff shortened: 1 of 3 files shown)


mdelapenya (approved):

adam-stokes (approved):

graphaelli (approved):
One nit if you want to address it - the other `$lang.sh` options are ordered:
```
DEFAULT_COMPOSE_ARGS="${ELASTIC_STACK_VERSION} ${BUILD_OPTS} \
  --no-apm-server-dashboards \
  --no-apm-server-self-instrument \
  --no-kibana \
  --with-agent-$lang \
  --force-build \
  --no-xpack-secure \
  --apm-log-level=debug"
```

while these are slightly different

PR #1247, opened 2021-09-03
Wait time for services to settle in EC

cachedout:
## What does this PR do?

Inserts a 60s wait time prior to starting tests as discussed with @axw and @kuisathaverat 

## Why is it important?

Trying to see if there might be an issue with EC wherein services aren't fully ready after they report themselves as being available.

### DIFF
--- a/.ci/integrationTestEC.groovy
+++ b/.ci/integrationTestEC.groovy
@@ -88,6 +88,11 @@ pipeline {
                   stash allowEmpty: true, includes: "${EC_DIR}/ansible/build/config_secrets.yml", name: "secrets-${getElasticStackVersion()}"
                 }
               }
+              stage("Waiting for services to settle") {
+                  steps {
+                      sleep 60
+                  }
+              }
               stage("Test Go") {
                 steps {
                   runTest('go')


axw (approved):
thanks!

.ci/integrationTestEC.groovy
```diff
@@ -88,6 +88,11 @@ pipeline {
                   stash allowEmpty: true, includes: "${EC_DIR}/ansible/build/config_secrets.yml", name: "secrets-${getElasticStackVersion()}"
                 }
               }
+              stage("Waiting for services to settle") {
+                  steps {
+                      sleep 60
```
mdelapenya, line 93: As a follow-up, we could make this value configurable as a jenkins input parameter, so that we can play with that value until getting the most accurate one.

mdelapenya (approved):

.ci/integrationTestEC.groovy
cachedout: The reason I didn't do that is that I want to be able to easily correlate between wait times and test results. Making it parametrized complicates that IMHO.

kuisathaverat (approved):

v1v (approved):

Issue #1241, opened 2021-08-31
Decrease sampling rate and drop unsampled transactions


sqren:
Sometime in the future APM Server will drop unsampled transactions by default. This will impact the APM ui, and to prepare ourselves for this shift I suggest that apm-integration-testing by default will drop unsampled transactions also. For this to have any noticable effect we should also reduce the sampling rate from 100% to something else. Perhaps 10% or maybe even 1%?

To drop unsampled transactions set `keep_unsampled: false` in APM Server:
https://www.elastic.co/guide/en/apm/server/master/transaction-metrics.html#sampling-keep_unsampled

The easiest way to change the sampling rate for all agents might be via [central configuration API](https://www.elastic.co/guide/en/kibana/current/agent-config-api.html). This would also ensure that we actually use central config on a daily basis.

cachedout:
Hi @sqren 

Thanks for the heads-up here. I'm also going to cc: @graphaelli because I think he might have some input.

We can certainly add this setting to the APM integration test suite by default if that's what the teams want to do. (That's quite easy.)

If we also want to change the same rate for each agent, we just need guidance on specifically what you want that value to be.

I think the next step here is to see what @graphaelli says and maybe we'll need to bring some agent teams in as well to figure out how far we want to decrease the sample rates on the agent side. Once all those decisions are made, a PR should be no problem.

Lastly, is this also something you want to happen on the observability test clusters? If so, we'll need to open a separate issue to track that but we can wait until after we've made decisions on the above.

sqren:
> If we also want to change the same rate for each agent, we just need guidance on specifically what you want that value to be.

Let's say 10% but I'm okay with anything below 10% really.

sqren:
> Lastly, is this also something you want to happen on the observability test clusters? If so, we'll need to open a separate issue to track that but we can wait until after we've made decisions on the above.

Yes, I think it would be great to have on the obs clusters as well. But fine to focus on apm-integration-testing first, and then tackle the obs clusters later.

graphaelli:
I think this makes a lot of sense to make it simple to drop unsampled tranasaction in apm-it, making it the default for 7.16+ clusters in apm-it with an option to keep unsampled sounds like a good path forward.  

[...]

PR #1255, opened 2021-09-10
Sampling rate overhaul

cachedout:
## What does this PR do?

1. Introduces a new CLI flag called `--drop-unsampled`. When this flag is passed in, `apm-server.sampling.keep_unsampled` is set to `true` on the APM server.
2. Introduces new flags for each Opbean (`--opbeans-$lang-sample-rate`) which controls the sampling rate for each opbean.
3. Various small fixups encountered along the way.
## Why is it important?
Requested by APM team.

## Related issues
Closes https://github.com/elastic/apm-integration-testing/issues/1241


docker/dyno/tests/conftest.py
```diff
@@ -19,7 +19,10 @@
 """
 Tests for Dyno control plane
 """
-import toxiproxy
+try:
+    import toxiproxy
```
cachedout, line 23: This just cleans up some errors that can happen with pytest during test discovery.

sqren:
This is awesome, thanks! Can we make it the default to drop unsampled, and have a sampling rate less than 100%?

cachedout:
Hi @sqren 

Here is the logic for deciding how to set the sampling flag: https://github.com/elastic/apm-integration-testing/pull/1255/files#diff-df11512a4602ab3dcd2e68f8f85f7462684d54050bfcb70ba7f94b39d53bea8aR203

The idea here was to set this as the default if it was set explicitly or the version running is at least 7.16 [as suggested here](https://github.com/elastic/apm-integration-testing/issues/1241#issuecomment-909627150).

Are you saying you'd like dropping un-sampled transactions to be the default behavior regardless of version?

sqren:
> The idea here was to set this as the default if it was set explicitly or the version running is at least 7.16 as suggested here.

That sounds great, thanks. I missed the point about 7.16.

So running `./scripts/compose.py start master` will drop unsampled transactions and have a sampling rate of 10%?

cachedout:
> @sqren  So running ./scripts/compose.py start master will drop unsampled transactions and have a sampling rate of 10%?

That is correct.

sqren (approved):
lgtm

mdelapenya (approved):

PR #1259, opened 2021-09-13
Use multi-stage Docker build instead of cURL

cachedout:
## What does this PR do?

This simplifies the Dockerfile used by the APM Integration Test suite to use a Docker multi-stage build instead of a more error-prone cURL-based approach for installing NodeJS.

## Why is it important?

We occasionally see failures [such as this one](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-integration-test-downstream/detail/master/10814/pipeline) in which the script-based approach for installing NodeJS fails. By using multi-stage builds, the hope is that we'll avoid those types of issues.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,7 @@
-FROM python:3.7
+FROM node:12-buster
+RUN npm install elasticdump -g
 
-#Install elasticdump
-RUN curl -sL https://deb.nodesource.com/setup_12.x | bash - \
-    && apt-get -yqq update \
-    && apt-get install -yqq nodejs \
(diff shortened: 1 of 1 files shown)


v1v (commented):
IIUC, `elasticdump` won't be available in the final docker image unless the `binary` is copied from the `node:12-buster` docker stage.

`elasticdump` is required as described in https://github.com/elastic/apm-integration-testing/pull/432

cachedout:
@v1v You're absolutely right. I've adjusted the build and in my testing there is now a working copy of elasticdump in built containers.

v1v (approved):

Issue #1187, opened 2021-06-29
Infer kibana's default "host:port" when starting apm-server without kibana


sqren:
APM Server needs to connect to Kibana for certain features to work (eg. Central Configuration). 
By default APM Server expects Kibana to run on `localhost:5601`. However when running with `--no-kibana` flag APM Server should no longer try to connect to Kibana on localhost (since it clearly won't be available) but instead on `host.docker.internal:5601` which is the hostname for Kibana running on the host machine (I think this address only works for Mac, and `172.17.0.1:5601` is needed for Linux).

I currently run:

```
./scripts/compose.py start master --no-kibana --apm-server-opt="apm-server.kibana.host=host.docker.internal:5601"
```

but I was hoping that I could avoid the last part and simply do:
```
./scripts/compose.py start master --no-kibana 
```

mdelapenya:
@sqren thanks for opening this ticket, could you add an estimated target date for this so we prioritise it accordingly?

kuisathaverat:
the URL `host.docker.internal` only works on macOS see https://github.com/docker/for-linux/issues/264, It is possible to pass the URL of Kibana to whatever you want with the flag `./scripts/compose.py start 7.13.2 --apm-server-kibana-url=http://host.docker.internal:5601`, this flag currently does not work if you pass `--no-kibana`. We should fix this flag to allow to pass the Kibana URL always, even when `--no-kibana` is passed.

sqren:
@kuisathaverat Maybe I was not clear in my description:

Can we automatically set the kibana host when running `--no-kibana`? The defaults should be
`--apm-server-opt="apm-server.kibana.host=host.docker.internal:5601"` on mac and `--apm-server-opt="apm-server.kibana.host=172.17.0.1:5601"` on linux.

This can obviously still be overridden but more sensible defaults than `apm-server.kibana.host=localhost:5601` like it is currently.

@mdelapenya There's no rush. Just an enhancement request.

sqren:
Please let me know if this is still not clear.

cachedout:
@sqren Could you please checkout https://github.com/elastic/apm-integration-testing/pull/1263 and see if it works as you would expect? Thanks!

cachedout:
After some review, it looks like there is [a fundamental problem](https://github.com/elastic/apm-integration-testing/pull/1263#issuecomment-919056808) with this approach. Unless there is a pressing need, I think that we should close this as "won't fix".

[...]

PR #1263, opened 2021-09-14
Infer default host:port when starting APM Server without Kibana

cachedout:
## What does this PR do?

When `--no-kibana` is passed in, APM Server is configured to select the correct platform-specific URL for Kibana.

## Why is it important?

Requested by @sqren 

## Related issues
Closes https://github.com/elastic/apm-integration-testing/issues/1187

### DIFF
--- a/scripts/modules/elastic_stack.py
+++ b/scripts/modules/elastic_stack.py
@@ -4,6 +4,7 @@
 
 import argparse
 import json
+import platform
 import os
 
 from .helpers import curl_healthcheck, try_to_set_slowlog, urlparse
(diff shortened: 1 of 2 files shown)


kuisathaverat:
I think this will cause issues, with these changes now if you pass `--no-kibana` the environment variables that point to Kibana will be configured, in case you do not have a Kibana running anything that tries to push dashboards, ingest pipelines, and other Kibana configuration will fail. We run several tests with the flag `--no-kibana` to avoid push those objects.

cachedout:
> We run several tests with the flag --no-kibana to avoid push those objects.

If that is the case, then I agree that we shouldn't do this.

sqren:
This looks great! Thanks!

sqren (approved):

cachedout:
Upstream issue closed as WONTFIX.

PR #1264, opened 2021-09-14
Upgrade urllib3, py, and requests

cachedout:
## What does this PR do?

* Upgrades urllib3 to avoid CVE-2020-26137
* Upgrades py to avoid CVE-2020-29651
* Upgrades requests (due to urllib3 upgrade)

## Why is it important?

This isn't a particularly urgent security concern. Just keeping things tidy. :)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,7 +17,7 @@ idna==2.6
 jsonschema==2.6.0
 mccabe==0.6.1
 pluggy==0.13.1
-py==1.5.2
+py==1.10.0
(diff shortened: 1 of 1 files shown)


kuisathaverat (approved):

beniwohli (approved):
`py` is a dependency of `pytest`, which is now at version 6.2.5. Might make sense to update that (and all the pytest dependencies) soon, too.

v1v (approved):

PR #1269, opened 2021-09-15
fix: Remove `UseAVX=2` flag on ARM64 machines

marclop:
## What does this PR do?

Adds another check in the `if` clause that adds the `UseAVX=2` flag to
the Elasticsearch container to not add it when the machine CPU arch is
ARM64. It should work for both Linux and macOS.

## Why is it important?

Allow engineers with an ARM workstation to run the tests.

### DIFF
--- a/scripts/modules/elastic_stack.py
+++ b/scripts/modules/elastic_stack.py
@@ -5,6 +5,7 @@
 import argparse
 import json
 import os
+import platform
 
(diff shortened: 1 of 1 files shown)


kuisathaverat (approved):
LGTM

simitt (approved):
🎉

PR #1280, opened 2021-10-06
Add missing sleep backport to ES check

cachedout:
## What does this PR do?

In the 7.x branch and the master branch, we have a sleep in between the retries. In the 6.x branch, that change was never applied. I believe this may be the root cause of a l[arge number of errors on the 6.x branch](https://jenkins-stats-oblt.elastic.dev/goto/fa14b837e0779427ad6c8fc39219ac73) which do not occur in 7.x and master.

## Why is it important?

Hopefully increased test stability on 6.x branch

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -1,4 +1,5 @@
 from timeout_decorator import TimeoutError
+import time
 import requests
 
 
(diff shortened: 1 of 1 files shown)


v1v (approved):
I wonder whether we should automatically create backports to `6.x`

kuisathaverat (approved):

PR #1312, opened 2021-11-11
Adapt to breaking changes in 8.0

axw:
## What does this PR do?

- Jaeger gRPC is now always on, with no config, on the standard 8200 port.
- Secret token and API Key auth configuration names have changed.
- There is no longer an option to keep unsampled transactions.
- For self-instrumentation, the `apm-server.instrumentaton` config is no longer supported; we now use the libbeat `instrumentation` config.
- ILM configuration and index suffix configuration are no longer supported.
- Agent configuration cannot be disabled.
- Template setup via apm-server is no longer supported.
- Pipeline setup via apm-server is no longer supported.
- Remove the option to write to data streams; it is now enabled implicitly for 8.0+. Update tests to search data streams as well as legacy indices.

We now run a container to call the Fleet setup API when Kibana is enabled, in order to install the APM integration. Update Kibana healthcheck to wait for Kibana to be completely ready before calling the setup API.

Fleet requires security to be enabled, so we no longer disable security by default in CI.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ PHP_APACHE_URL ?= http://phpapacheapp
 RAILS_URL ?= http://railsapp:8020
 RUM_URL ?= http://rum:8000
 
-ES_USER ?= elastic
+ES_USER ?= admin
 ES_PASS ?= changeme
 ELASTIC_APM_SECRET_TOKEN ?= SuPeRsEcReT
 
--- a/scripts/modules/elastic_stack.py
+++ b/scripts/modules/elastic_stack.py
@@ -16,8 +16,6 @@ class ApmServer(StackService, Service):
 
     SERVICE_PORT = "8200"
     DEFAULT_MONITOR_PORT = "6060"
-    DEFAULT_JAEGER_HTTP_PORT = "14268"
-    DEFAULT_JAEGER_GRPC_PORT = "14250"
     DEFAULT_OUTPUT = "elasticsearch"
     OUTPUTS = {"elasticsearch", "file", "kafka", "logstash"}
     DEFAULT_KIBANA_HOST = "kibana:5601"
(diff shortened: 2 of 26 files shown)


axw:
The `fleet_setup` container should be removed when https://github.com/elastic/kibana/issues/111858 is implemented.

scripts/modules/elastic_stack.py
```diff
@@ -520,22 +414,21 @@ def _content(self):
...
 
         command = ["apm-server", "-e", "--httpprof", ":{}".format(self.apm_server_monitor_port)] + command_args
         if self.options.get("apm_server_enable_debug"):
             command = command + ["-d", "*"]
 
+        healthcheck_path = "/" if self.at_least_version("6.5") else "/healthcheck"
+        healthcheck = curl_healthcheck(self.SERVICE_PORT, path=healthcheck_path)
+        if self.at_least_version("8.0"):
+            curl_command = "curl -s http://localhost:{}/ | grep -q 'publish_ready.*true'".format(self.SERVICE_PORT)
+            healthcheck["test"] = ["CMD-SHELL", curl_command]
```
marclop, line 430: We can check this from `7.16` onwards, correct?

scripts/modules/elastic_stack.py
```diff
@@ -1175,6 +1068,8 @@ def __init__(self, **options):
                 self.environment["XPACK_FLEET_REGISTRYURL"] = url
             if use_local_package_registry:
                 self.depends_on["package-registry"] = {"condition": "service_healthy"}
+            if self.at_least_version("8.0"):
+                self.environment["XPACK_FLEET_PACKAGES"] = '[{"name":"apm","version":"latest"}]'
```
marclop, line 1072: Do we want to use the package version that matches the stack version here?

marclop (approved):

scripts/modules/elastic_stack.py
axw: Ideally, yes. At the moment it's 8.0.0-dev1. That tag might change over time (apparently that's what we should have been doing when updating storage), so I don't think we can predict the correct version to use.

I think we can defer this until 8.0 is frozen, and then set the version to "8.0.0" for the 8.0.x releases. When the package is bundled with Kibana we should have a method of telling Kibana to install the bundled version.

scripts/modules/elastic_stack.py
axw: We _can_, but it's not necessary - we're only enabling data streams by default for 8.0+. The server is "publish ready" by default in 7.16.

Makefile
```diff
@@ -23,7 +23,7 @@ PHP_APACHE_URL ?= http://phpapacheapp
 RAILS_URL ?= http://railsapp:8020
 RUM_URL ?= http://rum:8000
 
-ES_USER ?= elastic
+ES_USER ?= admin
```
simitt, line 26: Why does this need to change (did the `elastic` user not have any effect so far)?

scripts/modules/elastic_stack.py
simitt: I believe it is ok to test with the latest available patch version, and from what I understand the apmpackages >= 8.1 should not match if we set the minimum version in the package accordingly. So it would always fetch the latest available 8.0 package then.

simitt (approved):

Makefile
axw: AFAICT, there is no `elastic` user and CI has only been passing because security has been disabled. We need security now (for Fleet), and when turning that on I noticed the tests failed because they couldn't auth with elastic:changeme.

The container has an `admin` user seeded: https://github.com/elastic/apm-integration-testing/blob/master/docker/elasticsearch/users

axw:
@elastic/observablt-robots I'm merging to unblock some apm-server work, but I'd still appreciate another review to make sure I haven't broken any CI/automation-related things.

PR #1331, opened 2021-12-02
feat: enable pytest_otel plugin

kuisathaverat:
## What does this PR do?

* It adds the use of [pytest_otel](https://pypi.org/project/pytest-otel/) to report OpenTelemetry data 

## Why is it important?

It completes the OpenTelemetry report, so now we have the Jenkins spans plus the pyTest spans.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -160,6 +160,11 @@ dockerized-test: ## Run all the dockerized tests
 		-e ES_USER=$${ES_USER} \
 		-e ES_PASS=$${ES_PASS} \
 		-e ELASTIC_APM_SECRET_TOKEN=$${ELASTIC_APM_SECRET_TOKEN} \
+		-e OTEL_EXPORTER_OTLP_ENDPOINT \
+		-e OTEL_EXPORTER_OTLP_HEADERS \
+		-e OTEL_SERVICE_NAME \
+		-e TRACEPARENT \
+		-e OTEL_EXPORTER_OTLP_INSECURE \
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -160,6 +160,11 @@ dockerized-test: ## Run all the dockerized tests
 		-e ES_USER=$${ES_USER} \
 		-e ES_PASS=$${ES_PASS} \
 		-e ELASTIC_APM_SECRET_TOKEN=$${ELASTIC_APM_SECRET_TOKEN} \
+		-e OTEL_EXPORTER_OTLP_ENDPOINT \
+		-e OTEL_EXPORTER_OTLP_HEADERS \
+		-e OTEL_SERVICE_NAME \
```
v1v, line 165: IIRC, this service name is not defined in the `withOtelEnv`

- See https://github.com/elastic/apm-pipeline-library/tree/master/vars#withotelenv

The ones defined by the OTEL plugin are:

- https://github.com/jenkinsci/opentelemetry-plugin/blob/d3a89fd1e3f87103cc140499262fd8e42df8b283/src/main/java/io/jenkins/plugins/opentelemetry/JenkinsOpenTelemetryPluginConfiguration.java#L233-L254

I'll probably raise a PR in the apm-pipeline-library to set those env variable if version < 0.19

kuisathaverat:
```suggestion
		-e OTEL_SERVICE_NAME="apm-integration-testing" \
```

v1v (approved):

PR #1342, opened 2021-12-17
fix: set the correct APM Server environment variable for the RUM agent

kuisathaverat:
## What does this PR do?

* replaces `ELASTIC_APM_SERVER_URL` env var with `APM_SERVER_URL`

## Why is it important?

The RUM agent check the `APM_SERVER_URL` so because we are passing `ELASTIC_APM_SERVER_URL` the ESS test fails on ESS where the URL is not the default one `https://localhost:8200`. After this change the APM ITs on ESS should pass.


v1v (approved):
💯 it probably was hard to find this !

cachedout (approved):

PR #1345, opened 2021-12-20
mergify: keep the house clean and notify smarter

v1v:
Delete upstream branches when they are merged or closed.
Notify conflict only for contributors who are humans.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -54,11 +54,14 @@ pull_request_rules:
       queue:
         name: auto-merge
         method: squash
-  - name: delete upstream branch after merging changes on dev-tools/integration/.env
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -54,11 +57,14 @@ pull_request_rules:
...
     conditions:
-      - merged
-      - label=automation
-      - head~=^update-stack-version
+      - or:
+        - merged
+        - closed
+      - and:
+        - label=automation
+        - head~=^update-.*
```
v1v, line 67: This is the way we can monitor also the go changes

kuisathaverat (approved):

PR #1363, opened 2022-01-04
bump automation: use 7.17 branch

v1v:
## What does this PR do?

Use `7.17` since `7.x` is not a thing anymore

## Why is it important?

Otherwise it won't work the automation

### DIFF
--- a/.ci/bump-go-release-version.sh
+++ b/.ci/bump-go-release-version.sh
@@ -16,7 +16,7 @@ else
 fi
 
 ### Queries to the 7.x branch
-GO_VERSION=$(curl -s https://raw.githubusercontent.com/elastic/apm-server/7.x/.go-version)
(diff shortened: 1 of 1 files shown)


.ci/bump-go-release-version.sh
```diff
@@ -16,7 +16,7 @@ else
 fi
 
 ### Queries to the 7.x branch
-GO_VERSION=$(curl -s https://raw.githubusercontent.com/elastic/apm-server/7.x/.go-version)
+GO_VERSION=$(curl -s https://raw.githubusercontent.com/elastic/apm-server/7.17/.go-version)
```
v1v, line 19: The last 7 minor version is `7.17` therefore we can hardcode it here

kuisathaverat (approved):

PR #1460, opened 2022-03-24
fix: use stack version from the APM library

kuisathaverat:
## What does this PR do?

* It changes to use the Elastic Stack version defined in the APM pipeline Library

## Why is it important?

The versions rotate with every new release, this rotation causes the pipeline to fail because some Docker images are no longer available. The change in this PR used the version we have configured in the APM pipeline library, these versions rotate with every new release.

### DIFF
--- a/.ci/integrationTestEC.groovy
+++ b/.ci/integrationTestEC.groovy
@@ -2,7 +2,7 @@
 @Library('apm@current') _
 
 pipeline {
-  agent { label 'linux && immutable' }
+  agent { label 'ubuntu-20 && immutable' }
   environment {
     BASE_DIR="src/github.com/elastic/apm-integration-testing"
     EC_DIR="src/github.com/elastic/observability-test-environments"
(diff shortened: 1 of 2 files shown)


.ci/integrationTestEC.groovy
```diff
@@ -202,7 +83,43 @@ pipeline {
...
+    'go',
+    'dotnet',
+    'java',
+    'nodejs',
+    'php',
+    'python',
+    'ruby',
+    'rum',
+    'all'
+  ]
```
v1v, line 97: What do you think to use https://github.com/elastic/apm-pipeline-library/blob/main/vars/agentMapping.groovy ?

it requires to support a new step to list all the APM agents anyway

cachedout (approved):

.ci/integrationTestEC.groovy
kuisathaverat: none of those has the same list, it is a list on two places, we can move it but in a follow up.

v1v (approved):

PR #1512, opened 2022-08-11
Docker image version check

cachedout:
## What does this PR do?

⚠️ Should not be merged until we have also bumped the Docker images in the APM Integration Test suite to the most recent versions.

Adds a new test framework which can detect out-of-date Docker images.

Produces junit-xml output which can be consumed by Jenkins.

Screenshot 2022-08-12 at 16 26 02

## Why is it important?

This will allow us to pin all images but also be alerted when we need to upgrade one of them.

## Related issues
Closes https://github.com/elastic/observability-robots/issues/985


kuisathaverat:
I think we can use https://www.updatecli.io to make the same, we are using it in the oblt clusters to manage Docker image updates.

PR #1524, opened 2022-09-12
Remove integration tests

### LINKED ISSUE #1523: Remove agent tests
The agent integration tests are of limited value. The tests are fairly basic, and we have since added fake APM Server implementations to the agent repos, which validate payloads.

The cost of maintaining the integration tests exceeds the value they are providing, therefore we should remove them. Later on we might replace them with something a little more maintainable. For example: disentangling stack/server provisioning, and distributing the tests to each agent repository. We can come back to that as needed.

axw:
## What does this PR do?

- The server team stopped using the server integration tests a while ago, but never removed them. They are removed now.
- The server team has its own smoke tests for Elastic Cloud deployments. This includes upgrading from 7.17.x to 8.x, so remove the upgrade tests.
- Agent integration tests have not proved useful lately, and the maintenance cost is too high, so they are remove.d
- The Kibana integration tests are not run from any CI job, so they are removed.

As neither the server nor agents will rely on the integration tests any longer, the downstream jobs are removed. Only the main Jenkinsfile remains, which runs the unit tests for compose.py.

## Why is it important?

The cost of maintaining the integration tests exceeds the value they are providing.

## Related issues

Closes #1523


.ci/scripts/shellcheck
```diff
@@ -1,3 +0,0 @@
-#!/usr/bin/env bash
```
v1v, line 1: This might be required, since the `pre-commit` will use it

https://github.com/elastic/apm-integration-testing/blob/da2f61a7b29131b6a436f6eb8bd437878f67672b/.pre-commit-config.yaml#L18-L25

v1v (approved):
LGTM, just a minor comment for one script regarding the `pre-commit` that it's required

.ci/scripts/shellcheck
axw: That `pre-commit` hook is external, how does it reference this script?

v1v: IIRC, the `PATH` is set to point to `.ci/scripts` so it has a higher precedency. For vary reasons:

* We didn't want to depend on installing tools in the worker
* Avoid any security issues by fixing what version to be used since those are docker images we manage.
* Avoid to installing tools during the CI build if possible.

https://github.com/elastic/apm-integration-testing/blob/da2f61a7b29131b6a436f6eb8bd437878f67672b/.ci/Jenkinsfile#L91 is calling `preCommit` and https://github.com/elastic/apm-pipeline-library/blob/4bf7553413212e90a142c649ebb2ebac7b4a5503/vars/preCommit.groovy#L70-L75 is the 'magic' to reset the PATH env variable and use those tools.


A bit opinionated and likely hidden though

axw: I see, thanks! I'll replace the file and add a comment. I only removed it because I thought it was dead code.

axw:
Just waiting for https://github.com/elastic/apm-agent-java/pull/2802

v1v (approved):

PR #1543, opened 2023-01-17
action: use opentelemetry composite action

v1v:
## Motivation/summary

Back in the days we built the Jenkins OTEL integration, so let's use the same approach for the GH actions, so every workflow and step will be tracked as traces/spans.

## How to test these changes

See traces in the Elastic deployment

This has been already tested for quite sometime in some other projects

image

https://github.com/elastic/apm-pipeline-library/blob/main/.github/workflows/opentelemetry.yml


reakaleek (approved):

v1v:
@mergifyio update

